PR #78, opened 2015-03-14
Add 'toponly' parameter for recentchanges.

itkach:
When [rctoponly](http://www.mediawiki.org/wiki/API:Recentchanges) is set only changes which are the latest revision are listed, otherwise same titles are often included over and over again.


danmichaelo:
Thanks! Merged (fixed two PEP issues first, so I did a rebase) and added to [release notes](https://github.com/mwclient/mwclient/blob/master/RELEASE-NOTES.md)

PR #81, opened 2015-03-25
drop Page.section attribute: it does wrong things

AdamWill:
This attribute (I think) goes back to an older design, where
the 'edit page' workflow was different: the edit() method did
part of the work of defining what text you meant to edit. In
the current workflow, text() - which replaced the deprecated
edit() - is only for retrieving text, and save() is supposed
to do all the work of editing the page.

The 'section' attribute is therefore obsolete and dangerous,
because what it does is this: if you ever call text() and pass
the 'section' argument, then later call save() without passing
the 'section' argument, it assumes you want to save the same
section you retrieved with text(). That doesn't seem to be a
valid assumption with the expected and currently-documented
behaviour of text() and save(), and could lead to unexpected
and confusing results (for e.g. you intended to rewrite the
page entirely without reference to the contents of the section
you retrieved).

This change should be documented as a BC break, though, as it
is possible there are still users relying on the old design
and not passing 'section' when they call save().

### DIFF
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -15,7 +15,6 @@ class Page(object):
             return self.__dict__.update(name.__dict__)
         self.site = site
         self.name = name
-        self.section = None
 
         if not info:
             if extra_properties:
@@ -137,11 +136,9 @@ class Page(object):
         try:
             rev = revs.next()
             text = rev['*']
-            self.section = section
             self.last_rev_time = rev['timestamp']
         except StopIteration:
             text = u''
-            self.section = None
             self.last_rev_time = None
         if not expandtemplates:
             self.edit_time = time.gmtime()
@@ -160,9 +157,6 @@ class Page(object):
             raise errors.UserBlocked(self.site.blocked)
         if not self.can('edit'):
             raise errors.ProtectedPageError(self)
-
-        if not section:
-            section = self.section
 
         if not self.site.writeapi:
             raise errors.NoWriteApi(self)


AdamWill:
Hum, a thought - we could have the deprecated `edit()` behave in the old way (setting `self.section`) until it's actually removed, but not have `text()` do it?

danmichaelo:
I also thought along the lines of keeping the old behaviour in the deprecated `edit()` method..  Would you mind updating your pull request if you haven't changed your mind? A warning could be issued from the `save()` method by replacing
```python
        if not section:
            section = self.section
```
with something like
```python
        if self.section and not section:
            warnings.warn('From mwclient version 0.8.0, the `save()` method will no longer implicitly use ' +
                          'the `section` parameter from the last `text()` or `edit()` call. Please pass the ' +
                          '`section` parameter explicitly to the save() method to save only a single section.',
                          category=DeprecationWarning, stacklevel=2)
            section = self.section
```
I guess. And then in the next major version (0.8.0) we can probably remove all of this, including the `edit`method.

(Perhaps we should also make DeprecationWarnings visible by default, but I could add that in a later commit)

AdamWill:
OK, I'll update it in a bit. Thanks!

AdamWill:
Sorry, I sort of lost track of this with other work, I'll get back to it (and my caching thing) soon.

AdamWill:
@danmichaelo So I see you added the deprecation warning now: how would you like this re-done? Just re-based the way it is now, or would you like anything different? Thanks!

danmichaelo:
Hm, I'm not really sure what's the best way to do this, but since it was deprecated only ~ a month ago in 4d8abc405 , I think I'll prefer to just keep the issue open for some more time until I eventually merge it in at some point before the release of 0.8.0.

danmichaelo:
Never mind, I'll just merge it in now, to make #80 easier to handle

PR #89, opened 2015-07-05
Adding pageid, contentmodel, pagelanguage, restrictiontypes as attributes of Page

PierreSelim:
Those values are only accessible through the private dict _info. Making them available directly may be useful to mwclient users.

### DIFF
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -48,6 +48,7 @@ class Page(object):
         self.length = info.get('length')
         self.protection = dict([(i['type'], (i['level'], i['expiry'])) for i in info.get('protection', ()) if i])
         self.redirect = 'redirect' in info
+        self.pageid = info.get('pageid', None)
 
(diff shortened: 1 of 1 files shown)


danmichaelo:
Won't hurt :)

scholer:
Would it be relevant to have these as properties? I.e. is it ever gonna be a problem that the data is duplicated? Let's say that a user first sets self.restrictiontypes = <something> and then self._info['restrictiontypes'] = <something else>. Then suddenly self.restrictiontypes won't match the the value in self._info. My feeling is that since _info is private, and not really modified in page.py, it shouldn't be an issue. (And if it ever becomes an issue, we can always make the attribute a property.)

danmichaelo:
Most likely it won't be a problem. But I guess most or all Page attributes could really be read-only properties. Feel free to submit a separate issue or pull request for this if you think it's worth the effort (and a good idea).

PR #92, opened 2015-08-12
Removing unsued import

PierreSelim:
Removing the following import:
* from client.py: time, random, sys, weakref
* from sleep.py: random


danmichaelo:
Thanks!

PR #96, opened 2015-08-25
Wikibase client (read only)

PierreSelim:
I've been working on implementation of reading method of the wikibase API:
- wikibase_item() in the page module
- WikiBaseSite and entities() in the client module
- entity module. The Claim class is mainly an helper to access dictionnary returned by wbgetclaims or wbgetentities method from API

Example of how it works:

``` python
import mwclient

site = mwclient.Site(('https', 'en.wikipedia.org'))
article = site.Pages['Paris']
item = article.wikibase_item()

# wikibase repo
wikidata = site.wikibase()
print wikidata

# Claims
claims = item.claims()
for claim in claims:
    print claim
    print claim.value

# Only coordinates
for claim in item.claims(prop=['P625']):
    print "%f, %f" % (claim.latitude, claim.longitude)

# Only subdivision (Paris subdivision are Items)
for claim in item.claims(prop=['P150']):
    print claim.item

wikidata = mwclient.WikiBaseSite(('https', 'www.wikidata.org'))
entities = wikidata.entities(ids=['Q42', 'P238'])
for entity in entities:
    print entity
    if entity.entity == 'P238':
        for claim in entity.claims(prop=['P1659']):
            print claim.property_value
# <Property object P238 (('https', 'www.wikidata.org'))>
# <Property object P239 (('https', 'www.wikidata.org'))>
# <Property object P229 (('https', 'www.wikidata.org'))>
# <Item object Q42 (('https', 'www.wikidata.org'))>
```

### DIFF
--- a/mwclient/entity.py
+++ b/mwclient/entity.py
@@ -1,0 +1,277 @@
+"""WikiBase Entities and related objects."""
+
+
+class Entity(object):
+
+    """Wikibase Entity, either Item or Property.
+
+    This class should not be implemented directly,
+    It is meant to be abstract for Item and Property.
+
+    Attributes:
+        site (WikiBaseSite): reference to a WikiBaseSite
+        entity (str): Q number of the entity.
+        descriptions (dict): dictionary containing description per language
+        labels (dict): dictionary containing labels per language
+    """
+
+    def __init__(self, site, normalized_entity):
+        """Common part of constructor for Item and Property."""
+        self.site = site
+
+        self.entity = normalized_entity
+        # caching descriptions, labels, sitelinks
+        # self._sitelinks = None
+        self._descriptions = None
+        self._labels = None
+
+        # caching claims
+        self._itemclaims = None
+
+    def setinfofromwbgetentities(self, result):
+        """Set descriptions, labels and claims from wbgetentities result."""
+        self._descriptions = dict()
+        for language in result['descriptions']:
+            lang = result['descriptions'][language]['language']
+            value = result['descriptions'][language]['value']
+            self._descriptions[lang] = value
+        self._labels = dict()
+        for language in result['labels']:
+            lang = result['labels'][language]['language']
+            value = result['labels'][language]['value']
+            self._labels[lang] = value
+        if self._itemclaims is None:
+            self._itemclaims = []
+            for prop in result['claims']:
+                for claim in result['claims'][prop]:
+                    mainsnak = claim['mainsnak']
+                    self._itemclaims.append(Claim.fromsnak(self.site,
+                                                           mainsnak))
+
+    @property
+    def labels(self):
+        """Labels dictionary per language"""
+        if self._labels is None:
+            entities = self.site.api('wbgetentities', ids=self.entity)
+            result = entities['entities'][self.entity]
+            self.setinfofromwbgetentities(result)
+        return self._labels
+
+    @property
+    def descriptions(self):
+        """Descriptions dictionary per language"""
+        if self._descriptions is None:
+            entities = self.site.api('wbgetentities', ids=self.entity)
+            result = entities['entities'][self.entity]
+            self.setinfofromwbgetentities(result)
+        return self._descriptions
+
+    def claims(self, prop=None):
+        """Claims about an Entity.
+
+        API Doc: https://www.mediawiki.org/wiki/Wikibase/API/en#wbgetclaims
+        We will probably need to implement rank and props later on.
+
+        Args:
+            prop (list, optional): list of property e.g. ['P238', 'P239']
+        """
+        if self._itemclaims is None:
+            self._itemclaims = []
+            info = self.site.api('wbgetclaims', entity=self.entity)['claims']
+
+            for propid in info:
+                for claim in info[propid]:
+                    mainsnak = claim['mainsnak']
+                    self._itemclaims.append(Claim.fromsnak(self.site,
+                                                           mainsnak))
+        if prop is None:
+            return self._itemclaims
+        else:
+            return [claim for claim in self._itemclaims if claim.prop in prop]
+
+
+class Item(Entity):
+
+    """Wikibase Item.
+
+    Attributes:
+        site (WikiBaseSite): reference to a WikiBaseSite
+        entity (str): Q number of the entity.
+        sitelinks (dict): dictionary containing sitelinks per wiki
+        descriptions (dict): dictionary containing description per language
+        labels (dict): dictionary containing labels per language
+    """
+
+    def __init__(self, site, entity):
+        """Constructor.
+
+        Args:
+            site (WikiBaseSite): reference to a WikiBaseSite
+            entity (str): Q number of the entity.
+        """
+        # Normalizing entity name
+        super(Item, self).__init__(site, 'Q' + entity.upper().lstrip('Q'))
+
+        self._sitelinks = None
+
+    def setinfofromwbgetentities(self, result):
+        """Set sitelinks, descriptions, labels, claims from wbgetentities."""
+        super(Item, self).setinfofromwbgetentities(result)
+        self._sitelinks = dict()
+        for wiki in result['sitelinks']:
+            site = result['sitelinks'][wiki]['site']
+            title = result['sitelinks'][wiki]['title']
+            badges = result['sitelinks'][wiki]['badges']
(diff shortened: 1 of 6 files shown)


danmichaelo:
Thanks for pushing this! I'll try to read over the code in the weekend.

PierreSelim:
@danmichaelo do not hesitate to comment things that you don't like in my code :)

danmichaelo:
When doing tab completion on the Site object, it's a bit confusing to have both `wikibase` and `wikibase_site`. I'll suggest prefixing the latter with `_`. And btw. we should really start using properties in mwclient :) This seems like a nice place to start. How about this?

```python
    @property
    def wikibase_repo(self):
        """
        Get a reference to the associated Wikibase repository site.

        Returns:
            a WikiBaseSite object
        """
        if self._wikibase_repo is None:
            result = self.api('query', meta='wikibase')
            url = result['query']['wikibase']['repo']['url']
            host = url['base'].replace('//', '')
            path = url['scriptpath'] + "/"
            self._wikibase_repo = WikiBaseSite(('https', host),
                                              path=path,
                                              pool=self.connection)
        return self._wikibase_repo
```

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+        self.site = site
+
+        self.entity = normalized_entity
+        # caching descriptions, labels, sitelinks
+        # self._sitelinks = None
+        self._descriptions = None
+        self._labels = None
+
+        # caching claims
+        self.itemclaims = None
```
danmichaelo: Prefix this one with underscore as well?

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+    def descriptions(self):
+        """Descriptions dictionnary per language"""
+        if self._descriptions is None:
+            entities = self.site.api('wbgetentities', ids=self.entity)
+            result = entities['entities'][self.entity]
+            self.setinfofromwbgetentities(result)
+        return self._descriptions
+
+    def claims(self, prop=None, rank=None, props=None):
+        """Claims of an Entity.
```
danmichaelo: *about* an Entity

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+        if self.itemclaims is None:
+            self.itemclaims = []
+            for prop in result['claims']:
+                for claim in result['claims'][prop]:
+                    mainsnak = claim['mainsnak']
+                    self.itemclaims.append(Claim.fromsnak(self.site, mainsnak))
+
+    @property
+    def labels(self):
+        """Labels dictionnary per language"""
```
danmichaelo: "Get a dictio<s>n</s>nary of labels per language code"

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+        if self._labels is None:
+            entities = self.site.api('wbgetentities', ids=self.entity)
+            result = entities['entities'][self.entity]
+            self.setinfofromwbgetentities(result)
+        return self._labels
+
+    @property
+    def descriptions(self):
+        """Descriptions dictionnary per language"""
```
danmichaelo: "Get a dictio<s>n</s>nary of descriptions per language code"

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+    def fromsnak(cls, site, snak):
+        """Claim from snak dictionnary.
+
+        Args:
+            site (mwclient.WikiBaseSite): site
+            snak (dict): snak dictionnary
+        """
+        return cls(site, snak['property'],
+                   snak['datatype'],
+                   snak['datavalue']['value'])
```
danmichaelo: Beware that 'datavalue' is only present if `snaktype="value"`. Here's an example from Q2 where it's not:

```
                "mainsnak": {
                    "datatype": "time",
                    "property": "P576",
                    "snaktype": "somevalue"
                },
```

Snaktype can have the values `"value"`, `"somevalue"` or `"novalue"`.

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def __repr__(self):
+        """Representation."""
```
danmichaelo: This docstring is not really helpful, and I don't see the point in having docstrings for `__repr__` anyways

mwclient/entity.py
```diff
@@ -0,0 +1,271 @@
...
+        return "<Claim object %s [%s]>" % (self.prop, self.datatype)
+
+    @property
+    def string(self):
```
danmichaelo: Not sure if I like this interface with one method for each datatype case. Wouldn't it be better to have just *one* property called `value` that returns the value no matter what datatype? Perhaps in the form of a `DataValue` class (implenting __str__ and __unicode__ so you could still use it as a string).

PierreSelim: I was even wondering about using ```attrdict``` for that, but your right this interface is clearly not something I love that much :)

mwclient/entity.py
PierreSelim: That would be the logic thing to do +1 :)

mwclient/entity.py
PierreSelim: Nice catch ^_^

mwclient/entity.py
PierreSelim: Indeed ...

.idea/dictionaries/Pierre_Selim.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="Pierre-Selim">
```
danmichaelo: And please remove this file ;)

PierreSelim: I'm ashamed this got pushed ... I just rebased it ^_^

mwclient/entity.py
PierreSelim: I'll try to have a look at this one today or next week, because it looks like a real blocker right now.

danmichaelo: Cool! I guess `snaktype` could be stored on the Claim object? Not sure what's best to do with `Claim.value`.. perhaps just set it to None?

PierreSelim: That's probably the simplest thing to do having both:
* `Claim.snaktype`
* `Claim.value` set to `None` if `Claim.snaktype` is either `somevalue` or `novalue`

waldyrious:
@PierreSelim and @danmichaelo, since these things sometimes fall off our radar, I'm just pinging you in case there's something other than dev work (e.g. a comment to respond) stalling the discussion.

PierreSelim:
@waldyrious and @danmichaelo true this has totally fell off my radar, I'll probably re-read the review and try to adress what I can this week. Still annoys me to use pywikibot when I don't need to just to get few simple wikibase API results :)

danmichaelo:
Cool! I'll try to respond quickly if you post something. If you want to discuss anything more in detail, we could also do a [Gitter chat](https://gitter.im/mwclient/mwclient).

PierreSelim:
@danmichaelo right now I'm rebasing and squashing for readability purpose :)

mwclient/entity.py
PierreSelim: fixed!

mwclient/entity.py
PierreSelim: This design is weird indeed.

I think I'd rather go for something like
* having a `raw_value` argument with the the snackvalue in it
* having the `value` being `Property`, `Item`, `Coordinates`, `str`, etc. based on the `datatype`

What do you think @danmichaelo ?

AdamWill:
hey again! @waldyrious looped me into helping out with maintaining mwclient since dan is inactive ATM. I'm looking through old pull requests, obviously this one is pretty large and will take a minute to get familiar with, but if you're still interested, could you look at rebasing it? One thing I'd definitely personally want to request to merge this: tests! I'd hate to merge a major new feature with no tests. Do you think you could write some? Thanks again!

AdamWill:
so this has a couple of flake failures, and still no tests. if you could look at that, that would be great. thanks.

PierreSelim:
Thanks @AdamWill for the reviews, I'm a bit rusty on this project (I've been long time away). I'll try to address the issue. I've rebased my branch, I'm looking at flake and mypy for tonight.

PR #105, opened 2015-12-22
Possibility to turn off ssl certificate validation and proxy connection

tosher:
* Possibility to turn off ssl certificate validation (for self-signed certs on corp. wikies)
* Proxy connection support.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -55,7 +55,7 @@ class Site(object):
     def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
                  max_retries=25, wait_callback=lambda *x: None, clients_useragent=None,
                  max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None,
-                 **kwargs):
+                 requests=None):
         # Setup member variables
         self.host = host
         self.path = path
@@ -64,7 +64,7 @@ class Site(object):
         self.compress = compress
         self.max_lag = text_type(max_lag)
         self.force_login = force_login
-        self.requests = kwargs.get('requests', {})
+        self.requests = requests or {}
 
         if isinstance(httpauth, (list, tuple)):
             self.httpauth = HTTPBasicAuth(*httpauth)
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -54,7 +54,8 @@ class Site(object):
 
     def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
                  max_retries=25, wait_callback=lambda *x: None, clients_useragent=None,
-                 max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None, do_ssl_cert_verify=True):
+                 max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None,
+                 do_ssl_cert_verify=True, proxies=None):
         # Setup member variables
         self.host = host
         self.path = path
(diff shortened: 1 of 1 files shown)


danmichaelo:
Sorry for the delay on this! The argument name `do_ssl_cert_verify` is quite long. What do you think about renaming it to just `verify` to follow requests? Another idea would be to just pass `**kwargs` to requests.

scholer:
I'd be very happy if **kwargs is passed on to requests. 
In my research lab we are maintaining a custom version of @tosher 's Mediawiker plugin for Sublime, which again has a custom (old) version of mwclient, all so that we can inject custom headers (with a stored SessionID data) before making the initial connection. Yes, there are other ways to do this, using do_init=False, and then updating headers/etc after init, but this would just require more customizations to Mediawiker instead. If we can simply configure the requests Session with kwargs on init, that would make life easier. And thanks again, Dan, for making this library! :)

tosher:
It will be requests specified kwargs?
```python
class Site(object):
    def __init__(self, ..... **kwargs):
        self.kwargs = kwargs
        # ...
    # in raw_call
        stream = self.connection.post(fullurl, data=data, files=files, headers=headers, **self.kwargs)
```

danmichaelo:
:+1: @tosher . But please name it something like `self.connection_kwargs = kwargs`.

tosher:
done

eighthave:
This kind of change should really be accompanied by tests.  This is messing with the core security of this library, and getting it wrong could mean the certificate verification gets disabled for people who need it.

tosher:
* It's not default option for mwclient
* It's base functionality of requests lib

tosher:
Maybe, variant like this will be better?
* http://python-redmine.readthedocs.org/configuration.html#connection-options
Special parameter for requests library, not global kwargs..

danmichaelo:
> Maybe, variant like this will be better?
> * http://python-redmine.readthedocs.org/configuration.html#connection-options
>
> Special parameter for requests library, not global kwargs..

Agree, that's a better solution since we will never get a conflict with internal arguments. Would you mind updating the pull request? Then I can merge it right away. Sorry I forgot about this.

tosher:
Ok, i will..

tosher:
Travis is broken?

danmichaelo:
> Travis is broken?

Related to one of the deps, we can ignore that.

Being a little picky, I'm not sure if we need to introduce kwargs here.. I'd suggest just adding `requests=None` as a param to the method, and then do:

    self.requests = requests or {}

(since adding `requests={}` as a param is [not a good idea](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments))

Also, could you squash all the commits into one commit? Just do a `git push --force` afterwards

tosher:
* 'Code quality decreased by -0.63%'
    - `Too many local variables`

:) No problem, i'll remove kwargs. 

> `all the commits into one commit`

ok

waldyrious:
@danmichaelo for future reference, PRs can [now](https://github.com/blog/2141-squash-your-commits) be squashed at merge time by the maintainer, so there's no need for the awkward `push -f force` dance anymore :)

danmichaelo:
> @danmichaelo for future reference, PRs can [now](https://github.com/blog/2141-squash-your-commits) be squashed at merge time by the maintainer, so there's no need for the awkward `push -f force` dance anymore :)

Oooh, that's useful

tosher:
Done.
> no need for the awkward push -f force dance anymore :)

P.S. I'm skipped the dance :)

danmichaelo:
Thanks, no dance needed :)

Issue #106, opened 2016-01-18
Python 2 gives objects, python3 gives dictionaries


pfctdayelise:
I don't know if I'm missing something obvious, but under python3 I seem to get dictionaries of data back rather than nice Image or Category or Page objects.

```python
import pytest
import mwclient


site = mwclient.Site('en.wikipedia.org')
page = site.Pages['Marcos Baghdatis']


def test_images():
    images = list(page.images())
    image = images[0]
    assert type(image) == mwclient.image.Image


def test_categories():
    cats = list(page.categories())
    cat = cats[0]
    assert type(cat) == mwclient.listing.Category
```

I made a virtualenv with python2, these tests pass. Under python3 they fail. 

```
(aomp):~/workspace/aomp$ py.test test_version.py --tb=line
============================================= test session starts ==============================================
platform linux -- Python 3.4.3, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /home/workspace/aomp, inifile: 
collected 2 items 

test_version.py FF

=================================================== FAILURES ===================================================
/home/workspace/aomp/test_version.py:12: assert <class 'dict'> == <class 'mwclient.image.Image'>
/home/workspace/aomp/test_version.py:18: assert <class 'dict'> == <class 'mwclient.listing.Category'>
=========================================== 2 failed in 4.09 seconds ==========================================
```

tosher:
The same as in https://github.com/mwclient/mwclient/issues/99?
Comment at 1 Nov 2015..

waldyrious:
@tosher you can [link directly to a comment](https://github.com/mwclient/mwclient/issues/99#issuecomment-152850031) by clicking on the date :)

tosher:
Oh, thanks! :)
fixed..

pfctdayelise:
@tosher sounds like the same issue. I don't quite follow that other issue, is it resolved on some branch waiting to be merged? or what's the story?

tosher:
This error was catched in experimental branch with relative imports (i'm using it in my plugin for Sublime text). I'm fixed this error in me rep.
But i didn't know about existing it in master branch. This case confirms it :)

Possible fix: https://github.com/mwclient/mwclient/issues/99#issuecomment-154751657

AdamWill:
I am seeing this too, it's a deal-breaker for moving my mwclient-based code to py3. The specific case I'm hitting is iterating over pages 'in' a category:

    import mwclient
    import sys

[...]

PR #108, opened 2016-02-03
[#106] fix GeneratorList with Python 3 (add __next__)

AdamWill:
Credit for this fix goes to @tosher:
https://github.com/mwclient/mwclient/issues#issuecomment-154751657
I just figured out exactly what changes his comment suggested,
and poked around a bit to figure out what was going on and why
they're needed.

`GeneratorList` (and things deriving from it, e.g. `Category`)
was broken for iteration purposes in Python 3. In Python 3
the iterator protocol requires a `__next__` method, not a `next`
method as in Python 2. 8d0650c renamed `List`'s `next` method
to `__next__` and added a small wrapper `next` method to account
for this, but did not make the same change for `GeneratorList`.
So when you iterate over a `GeneratorList` with Python 3, it
winds up calling `List.__next__` (since `GeneratorList` inherits
from `List`), not `GeneratorList.next`. So we don't hit the bit
that turns the results into mwclient page objects, and just get
the raw dicts from the `List` method.

We also have to change a couple of direct calls to `List.next`
to call `List.__next__` instead. If we don't do this, then when
the call originates from a child class, things get messed up,
because `List.next` calls `self.__next__` - and that will wind
up calling the child class' `__next__`, not `List`'s `__next__`.

Confused yet? :)

### DIFF
--- a/mwclient/listing.py
+++ b/mwclient/listing.py
@@ -57,7 +57,7 @@ class List(object):
             if self.last:
                 raise StopIteration
             self.load_chunk()
-            return List.next(self, full=full)
+            return List.__next__(self, full=full)
 
     def next(self, full=False):
         """ For Python 2.x support """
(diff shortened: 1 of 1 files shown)


AdamWill:
there is no way I'm writing the listing tests just to make the coverage number go up, sorry, I didn't sign on for that kinda work :P

PR #114, opened 2016-04-20
add 'end' arg for PageList

AdamWill:
this is sometimes useful; for instance I have a use case for
getting all pages in a given namespace that are alphabetically
between two specific points (because their names are in a
predictable format that involves alphabetically-sorted dates).
'end' is added as the last kwarg to avoid breaking any current
uses that rely on the kwarg ordering (thanks dan for pointing
that out).

### DIFF
--- a/mwclient/listing.py
+++ b/mwclient/listing.py
@@ -181,7 +181,7 @@ class Category(mwclient.page.Page, GeneratorList):
 
 class PageList(GeneratorList):
 
-    def __init__(self, site, prefix=None, start=None, namespace=0, redirects='all'):
+    def __init__(self, site, prefix=None, start=None, namespace=0, redirects='all', end=None):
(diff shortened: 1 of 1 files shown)


danmichaelo:
Sorry about the delay! Agree that this is useful. It looks a bit strange to not have `end` after `start`, but yeah, it's good for backward compat :)

danmichaelo:
@AdamWill, would it be useful to add `end` to `Site.allpages()` (and the other `Site.all*` methods that do not have it) as well?

AdamWill:
yeah, probably - that's actually how I'm using it, just I have custom generators of the exact same type downstream:

https://www.happyassassin.net/cgit/wikitcms/tree/wikitcms/wiki.py#n161

so sure, seems like the right thing to do. Do you want to do it or should I?

PR #119, opened 2016-05-07
Rename constructor arg requests to reqs to Fix #118

### LINKED ISSUE #118: NoneType has no Session()
The merge of #105 seems to have broken unit tests (and the client ^_^)

```
>           self.connection = requests.Session()
E           AttributeError: 'NoneType' object has no attribute 'Session'

../official_mwclient/mwclient/mwclient/client.py:91: AttributeError
============================================================================ 20 failed, 21 passed in 1.24 seconds ============================================================================
```

I tried with the python interprer, I have the same results:
```
>>> import mwclient
>>> mwclient.__ver__
'0.8.2.dev1'
>>> mwclient.Site('commons.wikimedia.org')
  File "<stdin>", line 1, in <module>
  File "mwclient/client.py", line 91, in __init__
    self.connection = requests.Session()
AttributeError: 'NoneType' object has no attribute 'Session'
```

I'd say masking the requests package with an argument name is not a good idea ^_^

PierreSelim:
Beaware argument requests from the constructor has been rename reqs

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -55,7 +55,7 @@ class Site(object):
     def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
                  max_retries=25, wait_callback=lambda *x: None, clients_useragent=None,
                  max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None,
-                 requests=None):
+                 reqs=None):
         # Setup member variables
         self.host = host
         self.path = path
@@ -64,7 +64,7 @@ class Site(object):
         self.compress = compress
         self.max_lag = text_type(max_lag)
         self.force_login = force_login
-        self.requests = requests or {}
+        self.requests = reqs or {}
 
         if isinstance(httpauth, (list, tuple)):
             self.httpauth = HTTPBasicAuth(*httpauth)


danmichaelo:
Thanks :) Strangely the Travis build still fails, gotta figure out why..

scholer:
Hi,
Sorry to be commenting on such an old PR. But I just found myself looking for the "requests kwargs" parameter to send to ```Site.__init__```, and I completely missed that it was called ```reqs```. It's been a while since I last had to initialize a ```mwclient.Site``` directly, so it might just be me. But, it made me think whether it would make sense to rename the parameter to something slightly more verbose, e.g. ```requests_args```? Or alternatively document the ```reqs``` parameter it in the docstring of ```Site```?

There haven't been any official mwclient release since this PR, so I figure it's still OK to change parameter names if we want.

danmichaelo:
No prob, I actually agree when I look at it now. `reqs` isn't very intuitive, and it means autocompletion doesn't help you if you type something like `requ`.. so I'm open for changing it. Is `requests_args` what we settle for?

danmichaelo:
Ping @tosher as the original author of #105

tosher:
I'm agree with any variant :) 
.. in Mediawiker's version on client i'm using `kwargs` instead of sep. parameter :)

PR #121, opened 2016-05-08
make examples python-3 compatible

waldyrious:
This clears errors detected by landscape.io:
https://landscape.io/github/mwclient/mwclient/35/messages/error

Fixes based on recommendations here: http://python-future.org/compatible_idioms.html

### DIFF
--- a/examples/basic_edit.py
+++ b/examples/basic_edit.py
@@ -1,9 +1,10 @@
+from __future__ import print_function
 import sys, os
+
 if len(sys.argv) > 3:
 	sys.path.append(os.path.abspath(sys.argv[3]))
 if len(sys.argv) < 3:
-	print 'python basic_edit_test.py <config> <prefix> [<include path>]'
-	print
+	print('python basic_edit_test.py <config> <prefix> [<include path>]\n')
 	sys.exit()
 
 ## Create a config file containing:
@@ -12,7 +13,6 @@ if len(sys.argv) < 3:
 # ext = '.php'
 # username = 'Bryan'
 # password = 'xyz'
-
 
 prefix = sys.argv[2]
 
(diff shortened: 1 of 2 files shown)


waldyrious:
Landscape.io report for this change: https://landscape.io/diff/343642 (4% improvement)

waldyrious:
@danmichaelo are the ["smells" detected by landscape.io](https://landscape.io/github/mwclient/mwclient/39/messages/smell) accurate? I would fix those but wasn't sure whether some of those choices were deliberate.

danmichaelo:
> @danmichaelo are the ["smells" detected by landscape.io](https://landscape.io/github/mwclient/mwclient/39/messages/smell) accurate? I would fix those but wasn't sure whether some of those choices were deliberate.

The `Unused arguments` smells were quite bad.. I fixed those: 524d751318ee00ac11250c549982dc2d998a771f and a148d30629a3fc1a3003b3929e971a8ba0dd90f8

Also fixed two minor smells and ignore two others: 50bc28323e2536e58f591879799a35729daf7b2b

danmichaelo:
I'm keeping the cyclic-import warning, since it would be nice to eventually get rid of that, but it's not trivial.

danmichaelo:
Seems like `# pylint: disable=unused-import` didn't work... oh, well

waldyrious:
👍 thanks for taking a look. 98% code health is quite good!

waldyrious:
Btw, what about these?

- 'mwclient.errors.*' imported but unused (__init__.py:26)
- 'mwclient.client.Site' imported but unused (__init__.py:27)
- Arguments number differs from overridden '\__next__' method (listing.py:142)
- Arguments number differs from overridden 'next' method (listing.py:150)

danmichaelo:
The unused imports are used to export the classes (so you can do `from mwclient import Site`, not `from mwclient.client import Site`. I didn't want to disable unused-imports check completely, so I just added `# pylint: disable=unused-import` on the relevant imports, but seems like it didn't work.

"Arguments number differs" would be worth fixing, but I couldn't decide on how. The easiest way would be to add the `full` argument to the child methods, since that certainly wouldn't break anything, but extra arguments on `next()` is weird, so I would prefer to remove it altogether… but then I need to check that it doesn't break anything…

waldyrious:
Glad your thinking got documented :) hopefully these will eventually be fixed one way or another.

Issue #87, opened 2015-06-26
Add 'end' argument for various lists/generators


AdamWill:
Quite a lot of lists and generators have a `start` argument letting you specify where the results should start, but none of them seem to have the obvious corresponding `end` argument letting you specify where the results should stop. The API generators do offer this - whenever the `allpages` generator is being used, for instance, you can specify `gapto` (as you specify `gapfrom` for the `start` argument). For instance, to handle it in `PageList` the diff would just be:

    diff --git a/mwclient/listing.py b/mwclient/listing.py
    index 0e68e94..611221c 100644
    --- a/mwclient/listing.py
    +++ b/mwclient/listing.py
    @@ -161,7 +161,7 @@ class Category(page.Page, GeneratorList):
     
     class PageList(GeneratorList):
     
    -    def __init__(self, site, prefix=None, start=None, namespace=0, redirects='all'):
    +    def __init__(self, site, prefix=None, start=None, end=None, namespace=0, redirects='all'):
             self.namespace = namespace
     
             kwargs = {}
    @@ -169,6 +169,8 @@ class PageList(GeneratorList):
                 kwargs['gapprefix'] = prefix
             if start:
                 kwargs['gapfrom'] = start
    +        if end:
    +            kwargs['gapto'] = end
     
             GeneratorList.__init__(self, site, 'allpages', 'ap',
                                    gapnamespace=text_type(namespace), gapfilterredir=redirects, **kwargs)

danmichaelo:
Thanks for spotting this! If we introduce the `end` argument right after `start` (rather than at the end) it might break users depending on the order of arguments, but it makes much more sense, so I think we schedule it for 0.8 anyways.

AdamWill:
Sigh, so I actually was carrying this downstream, then took it out because I figured it had gone upstream and/or I wasn't using it any more, then today I find the thing that was using it and of course it doesn't work any more :sob: I guess I'll send a PR for the case I actually care about at least.

AdamWill:
https://github.com/mwclient/mwclient/pull/114 addresses this.

danmichaelo:
Thanks for the PR and sorry it took so long.

danmichaelo:
This is completed now, right?

AdamWill:
Sure!

PR #123, opened 2016-05-18
add 'end' argument for `allpages`, `allusers` etc. generators

AdamWill:
this follows on from #114 and addresses #87, providing an `end`
arg for all these generators, letting you get 'all pages between
Bar and Foo' or whatever it is you may need.

Untested, I just followed how the 'start' stuff works now.


danmichaelo:
Great! 🌟

PR #129, opened 2016-07-24
Use Super instead of explicit class naming

### LINKED ISSUE #127: Use `super`
Glancing over `listing.py`, I see a lot of explicit references of the parent class type, e.g.:
```python
class GeneratorList(List):

    def __init__(self, site, list_name, prefix, *args, **kwargs):
        List.__init__(self, site, list_name, prefix, *args, **kwargs)

    # …
```
In this case, `List` should be replaced by `super(GeneratorList, self)`. The same holds for every other occurrence.

lukasjuhrich:
Added some docstrings where appropriate as well.

Fixes #127

### DIFF
--- a/mwclient/errors.py
+++ b/mwclient/errors.py
@@ -19,7 +19,7 @@ class APIError(MwClientError):
     def __init__(self, code, info, kwargs):
         self.code = code
         self.info = info
-        MwClientError.__init__(self, code, info, kwargs)
+        super(APIError, self).__init__(code, info, kwargs)
 
 
 class InsufficientPermission(MwClientError):
@@ -75,7 +75,7 @@ class InvalidResponse(MwClientError):
                        'you used the correct hostname. If you did, the server might ' + \
                        'be wrongly configured or experiencing temporary problems.'
         self.response_text = response_text
-        MwClientError.__init__(self, self.message, response_text)
+        super(InvalidResponse, self).__init__(self.message, response_text)
 
     def __str__(self):
         return self.message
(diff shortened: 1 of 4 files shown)


lukasjuhrich:
Oh, I should probably take a look at the other files, too.

PierreSelim:
> Explicit is better than implicit.

Just saying :)

lukasjuhrich:
@PierreSelim I know that quite well.

But by the same argumentation you should never refactor out common constants because the that would be “implicit” and thus “bad”.

That's not the case here. The reason, at the positions I replaced it, the parent class has been used is because it was the parent class as given above. Therefore, it is quite explicit and to-the-point using a language construct with exactly that semantic meaning and use case.

I want the parent class. Which is not necessarily `List` in the future, but always what's been given above in the class definition.

This makes the code more reusable and easier changeable.

danmichaelo:
I think the reason super wasn't used was just for historical reasons, I don't think it was around in really old days. Thanks for fixing, this has been merged.

When it comes to docstrings, it seems like I hadn't documented it, but we sort-of settled on using the “Google style”, which isn't very strictly defined, but exemplified [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google). It's quite readable in the source code, and can also be parsed by the napoleon sphinx extension when producing html documentation (example: http://mwclient.readthedocs.io/en/latest/reference/site.html)

I added a note about it to the readme: https://github.com/mwclient/mwclient/commit/29ef67bd8a937fafabcfde10cbb69c980d9243a0

PR #132, opened 2016-08-11
cope with .ask() having more than 50 results

cariaso:
This uses query-continue-offset to retrieve all results when ask has more than 50 responses.
https://www.semantic-mediawiki.org/wiki/Ask_API#Limit_and_offset


danmichaelo:
Thanks! I rewrote the method a little bit before merging: https://github.com/mwclient/mwclient/commit/2bb79128d62441fe5d7ba607fba097ee9e8564db

PR #134, opened 2016-08-11
Drop python 2.6

### LINKED ISSUE #133: Drop python2.6 support
Some features which are quite common, such as the set literal, are only available in `python≥2.7`.

Because `python2.6` is officially discontinued by python and every major distribution uses 2.7, keeping this backwards compatibility does not seem worth any missing features.

Resolving this issue means effectively dropping 2.6 from the `.travis.yml` as well as from the PyPi distribution list.

@danmichaelo  answered to my comment on my pr (which was breaking the build due to using set literals) and said he was [open to dropping support](https://github.com/mwclient/mwclient/pull/131#issuecomment-238678420).

lukasjuhrich:
Fixes #133

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@
 sudo: false
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ mwclient
 
 mwclient is a lightweight Python client library to the `MediaWiki API <https://mediawiki.org/wiki/API>`_
 which provides access to most API functionality.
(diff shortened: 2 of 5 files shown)


lukasjuhrich:
…need to rebase, I didn't want it to include the changes of my other PR.

danmichaelo:
Merged in 59f79e82893c28c91788baee64b7f9b40683bb84, thanks!

AdamWill:
I'm very slightly sad about this as we ship mwclient in EPEL for RHEL 6, which we can't do any more if it doesn't support Python 2.6 any more (because RHEL 6's system python is 2.6 and always will be, I think). But only slightly sad as I really have no use cases for it myself.

PR #138, opened 2016-09-20
move badges to under the main heading

waldyrious:
otherwise the logo gets awkwardly separated from the title


danmichaelo:
12 minutes response time, this is good for http://issuestats.com/github/mwclient/mwclient ;)

PR #139, opened 2016-09-22
Enables SSL client certificate as authentication mechanism

nbareil:


danmichaelo:
Thanks! Would you mind also adding a short example under "Authenticating" in the user guide (
https://github.com/mwclient/mwclient/blob/master/docs/source/user/connecting.rst)

nbareil:
Sure thing, here it is. I am totally lost with reStructuredText syntax so maybe my links will not be correct.

PR #143, opened 2016-10-26
Support sending custom headers with all requests

bd808:
Allow sending custom headers (e.g. X-Wikimedia-Debug) with all requests
by passing a dict of header data in the Site constructor.


danmichaelo:
Thanks for including a test, and sorry it took so long to merge!

PR #145, opened 2016-12-02
Improve PEP-8 compatibility for examples

dereckson:
Examples should adhere to PEP-8 guidelines, as these examples
not only document how to use correctly the client, but also
will serve as a first exposition to Python for some contributors.

Reference: https://www.python.org/dev/peps/pep-0008/


waldyrious:
Thanks for doing this! This is a clear improvement, so I'll merge right away 👍

PR #146, opened 2017-01-26
support https://pypi.python.org/pypi/requests-cache via an…

cariaso:
pass an optional argument to the mwclient constructor
    cache_args={} 
and cacheable requests will be cached into sqllite. Mongo and redis backends, and custom expiration policies are supported via additional arguments documented at
https://requests-cache.readthedocs.io/en/latest/user_guide.html#persistence

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -11,6 +11,10 @@ try:
 except ImportError:
     import simplejson as json
 import requests
+try:
+    import requests_cache
+except ImportError, e:
+    requests_cache = None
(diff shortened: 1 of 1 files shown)


mwclient/client.py
```diff
@@ -11,6 +11,10 @@
 except ImportError:
     import simplejson as json
 import requests
+try:
+    import requests_cache
+except ImportError, e:
```
danmichaelo: This fails on Python 3. Please just use `except ImportError:`

PR #147, opened 2017-02-21
Send parse requests via POST

bd808:
In 1d61770 the Site.parse() command was changed from GET to POST. This
effectively limits the amount of wikitext that can be rendered to what
will fit in the URI limit of the webserver serving the MediaWiki site.
We could do something really tricky to look at the size of the text
argument and switch between GET and POST based on that, but it seems
much simpler to just stick with POST for this action.

Downstream bug: https://phabricator.wikimedia.org/T158715


danmichaelo:
Thanks! Agree that it's better to use POST with all requests here

Issue #153, opened 2017-05-06
Client:ask  broken ?


ubibene:
Hi all,

as it is , I cannot get Client:ask to work , 

For example, test it on Semantic Mediawiki Sandbox 

```
site=mwclient.Site(('https','sandbox.semantic-mediawiki.org'),'/w/')
query="[[Has email::%2B]]|%3FLabel" 
# see https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Parcourir/:Azure  : this page should be returned by ask
# + note that the arguments must be url encoded (see https://www.semantic-mediawiki.org/wiki/Ask_API )
res=site.ask(query)
res.next()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/mwclient/client.py", line 960, in ask
    for result in results['query']['results']:
```
The trouble is  : there is no "query" field in the response returned by site.ask() , the api calls returns an error

But you can see the query results here  : 

https://sandbox.semantic-mediawiki.org/w/api.php?action=ask&query=[[Has%20email::%2B]]|%3FLabel&format=json

 Cheers

PS :  what is the use of the Title parameter in Client.ask() ? 

MB

ubibene:
OK, in client.py

```
960            results = self.raw_api('ask', query='{query}|offset={offset}'.format(
961                     query=query, offset=offset, http_method='GET'), **kwargs)
```

should probably be

```
960            results = self.raw_api('ask', query='{query}|offset={offset}'.format(
961                     query=query, offset=offset**)** , http_method='GET', **kwargs)
```
But it still doesnt't work

ubibene:
Got it !
```
960            results = self.raw_api('ask', query='{query}|offset={offset}'.format(
961                     query=query, offset=offset) , http_method='GET', **kwargs)
```
should be

```
960            results = self.raw_api('ask', query='{query}&parameters=offset%3D{offset}'.format(
961                     query=query, offset=offset) , http_method='GET', **kwargs)
```

That is , use **'&parameters=offset%3D{offset}'** to conform to the API .

This gives :

```
>>> query
'[[Has email::%2B]]|%3FLabel'
>>> results = site.raw_api('ask', query='{query}&parameters=offset%3D{offset}'.format(query=query, offset=offset) , http_method='GET', **kwargs)
>>> results
[...]

PR #154, opened 2017-05-06
Issue #153 : fix semantic mediawiki api call

ubibene:
1.  Fix the closing parenthesis location  in ask() function
2. Fix the offset syntax to conform to Semantic Mediawiki API


ubibene:
Humm , something is wrong . What was working form calling self.raw_api from the command line , does not work when in Client.ask . I must have done something wrong. I'll have to check

Issue #161, opened 2017-05-11
Client.ask() fails if there is no anwser to the semantic query. or if the query is malformed


ubibene:
Damned !

In Client.ask()
```
results = self.raw_api('ask', query='{query}|offset={offset}'.format(
                query=query, offset=offset), http_method='GET', **kwargs)

            offset = results.get('query-continue-offset')
            for key, value in results['query']['results'].iteritems():
                yield {key: value}
```
If the semantic query gets results:
-  results['query ']['results'] is a dictionnary holding the answers , 
- otherwise it is an empty list and results['query']['results'].iteritems() breaks ( AttributeError: 'list' object has no attribute 'iteritems' )

My mistake. 

I'll look into it,

danmichaelo:
Note: I updated the generator to not yield an empty dictionary if there's zero results. To check if there's zero results, it's better to count the number of results you get from the generator

    answers = site.ask('..')
    num_answers = len(list(answers))

ubibene:
Thanks, it's cleaner that way. I'll post a message to the SMW user list about it.

PR #163, opened 2017-05-16
Fix issue #161 / from a brand new fork

ubibene:
I wasn't able to figure out how to fix the git mess.  

So I created a brand new fork form mwclient and added the fix for Issue  #161  to it. 

Hope this won't break anything.
Cheers
ubibene

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -969,7 +969,19 @@ class Site(object):
         API doc: https://semantic-mediawiki.org/wiki/Ask_API
 
         Returns:
-            Generator for retrieving all search results
+            Generator for retrieving all search results.
+            Iterating over the generator gives a dictionary containing the answers
+            to the query.
+            The dictionary is empty if the query is valid but there are no answers.
+
+        Examples:
+
+            >>> query = "[[Category:my cat]]|[[Has name::a name]]|?Has property"
+            >>> answer = site.ask(query)
+            >>> for a in answer:
+            >>>     for key, value in a.items()
+            >>>         print(key)
+            >>>         print(value)
         """
         kwargs = {}
         if title is None:
@@ -977,9 +989,14 @@ class Site(object):
 
         offset = 0
         while offset is not None:
-            results = self.raw_api('ask', query='{query}|offset={offset}'.format(
+            results = self.raw_api('ask', query=u'{query}|offset={offset}'.format(
                 query=query, offset=offset), http_method='GET', **kwargs)
-
-            offset = results.get('query-continue-offset')
-            for key, value in results['query']['results'].iteritems():
-                yield {key: value}
+            if self.handle_api_result(results):
+                answer = results['query'].get('results')
+                if answer:
+                    offset = results.get('query-continue-offset')
+                    for key, value in answer.items():
+                        yield {key: value}
+                else:  # query returns no results
+                    offset = None
+                    yield {}


danmichaelo:
Good idea :) Would you just mind changing "dictionnary" to "dictionary"? Then just do `git commit --amend` and `git push --force` (force pushing should be used carefully in general, but when working with PRs it's very useful to avoid ending up with lots of commits)

ubibene:
done !

mwclient/client.py
```diff
@@ -969,7 +969,21 @@ def ask(self, query, title=None):
         API doc: https://semantic-mediawiki.org/wiki/Ask_API
 
         Returns:
-            Generator for retrieving all search results
+            Generator for retrieving all search results.
+            Iterating over the generator gives a dictionnary containing the answers
+            to the query.
+            The dictionnary is empty if either:
+               - the query is valid but there are no answers,
+               - or if the query is not valid.
```
danmichaelo: This should probably raise an exception, see below

mwclient/client.py
```diff
@@ -969,7 +969,21 @@ def ask(self, query, title=None):
...
+
+        Examples:
+
+            >>> query = "[[Category:my cat]]|[[Has name::a name]]|?Has property"
+            >>> answer = site.ask(query)
+            >>> for a in answer:
+            >>>     for key,value in a.iteritems() # or a.items() in python 3
```
danmichaelo: To comply with PEP8 guidelines there should be a space after the comma (key, value) and two spaces before the comment sign (#)

danmichaelo: Btw. `a.items()` works in both Python 2 and 3, it just doesn't work as an iterator in Python 2, so for the sake of this example, I think it would be better to use `a.items()` and skip the comment.

mwclient/client.py
```diff
@@ -979,7 +993,15 @@ def ask(self, query, title=None):
         while offset is not None:
             results = self.raw_api('ask', query='{query}|offset={offset}'.format(
                 query=query, offset=offset), http_method='GET', **kwargs)
-
-            offset = results.get('query-continue-offset')
-            for key, value in results['query']['results'].iteritems():
-                yield {key: value}
+            if results.get('error'):  # malformed query, maybe should raise an exception
```
danmichaelo: Agree that this should raise an exception. Could you try passing results through `handle_api_result()`, which will raise an exception if the results contains an error?

ubibene:
Hi here your are.

To test

```
import mwclient


site=mwclient.Site(('https','sandbox.semantic-mediawiki.org'),'/w/')
query_with_results="[[Has email::+]]|?Label" 
query_with_no_results="[[Has email::xxxx]]|?Label" 
malformed_query="[[Has em"

answer=site.ask(query_with_results)
empty_answer=site.ask(query_with_no_results)
error_answer=site.ask(malformed_query)

print("-----------")
print("non empty answer")
for a in answer:
    for k,v in a.items():
        print(k)
        print(v)

print("-----------")
print("empty answer")
for a in empty_answer:
    for k,v in a.items():
        print(k)
        print(v)

print("-----------")
print("maformed query")
for a in error_answer:
    for k,v in a.items():
        print(k)
        print(v)
print("-----------")

```

gives
```

-----------
non empty answer
Issue/2093/1# 429eac66925d6bdd9dfecad0fc7e7ce7
{u'fulltext': u'Issue/2093/1# 429eac66925d6bdd9dfecad0fc7e7ce7', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Issue/2093/1#_429eac66925d6bdd9dfecad0fc7e7ce7'}
Lorem ipsum Export
{u'fulltext': u'Lorem ipsum Export', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum_Export'}
Issue/2093/1
{u'fulltext': u'Issue/2093/1', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Issue/2093/1'}
Issue/1855 (Lorem ipsum)
{u'fulltext': u'Issue/1855 (Lorem ipsum)', u'displaytitle': u'Lorem ipsum', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Issue/1855_(Lorem_ipsum)'}
Lorem ipsum# 0a872dd5b2e992230eac6aa27ffc0174
{u'fulltext': u'Lorem ipsum# 0a872dd5b2e992230eac6aa27ffc0174', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum#_0a872dd5b2e992230eac6aa27ffc0174'}
Azure
{u'fulltext': u'Azure', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': [{u'displaytitle': u'', u'fulltext': u'Azure', u'namespace': 0, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Azure', u'exists': u'1'}]}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Azure'}
Lorem ipsum
{u'fulltext': u'Lorem ipsum', u'displaytitle': u'', u'exists': u'1', u'namespace': 0, u'printouts': {u'Label': []}, u'fullurl': u'https://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum'}
-----------
empty answer
-----------
maformed query
{u'error': {u'query': [u'Certains \xab <nowiki>[[</nowiki> \xbb dans votre requ\xeate n\u2019ont pas \xe9t\xe9 clos par des \xab ]] \xbb correspondants.']}}
  File "/home/mbenard/testfixmwclient.py", line 29, in <module>
    for a in error_answer:
  File "/media/mbenard/MB-PRO/mwclient-fix/mwclient/mwclient/client.py", line 998, in ask
    if self.handle_api_result(results):
  File "/media/mbenard/MB-PRO/mwclient-fix/mwclient/mwclient/client.py", line 305, in handle_api_result
    if info['error']['code'] in {u'internal_api_error_DBConnectionError',
KeyError: 'code'

```

Note that in case the query is malformed Semantic Mediawiki returns in results  : 

`{u'error': {u'query': [u'Certains \xab <nowiki>[[</nowiki> \xbb dans votre requ\xeate n\u2019ont pas \xe9t\xe9 clos par des \xab ]] \xbb correspondants.']}}`

mwclient/client.py
```diff
@@ -979,7 +993,16 @@ def ask(self, query, title=None):
         while offset is not None:
             results = self.raw_api('ask', query='{query}|offset={offset}'.format(
                 query=query, offset=offset), http_method='GET', **kwargs)
-
-            offset = results.get('query-continue-offset')
-            for key, value in results['query']['results'].iteritems():
-                yield {key: value}
+            if results.get('error'):
```
danmichaelo: Since `self.handle_api_result(results)` will check for this, you don't need to do it here. You can replace this and the next three lines with just `self.handle_api_result(results)`

mwclient/client.py
```diff
@@ -979,7 +993,16 @@ def ask(self, query, title=None):
...
+                if self.handle_api_result(results):
+                    offset = None
+                    yield {}
+            elif results.get('query'):
```
danmichaelo: This line can also be removed. If there is no 'query' in the results, then something would be seriously wrong, but the current test would just push that under the carpet :)

mwclient/client.py
```diff
@@ -969,7 +969,21 @@ def ask(self, query, title=None):
         API doc: https://semantic-mediawiki.org/wiki/Ask_API
 
         Returns:
-            Generator for retrieving all search results
+            Generator for retrieving all search results.
+            Iterating over the generator gives a dictionary containing the answers
+            to the query.
+            The dictionary is empty if either:
+               - the query is valid but there are no answers,
+               - or if the query is not valid.
```
danmichaelo: Please update the comment to reflect that the dictionary is empty of if there is no answers

mwclient/client.py
```diff
@@ -979,7 +993,16 @@ def ask(self, query, title=None):
...
-            offset = results.get('query-continue-offset')
-            for key, value in results['query']['results'].iteritems():
-                yield {key: value}
+            if results.get('error'):
+                if self.handle_api_result(results):
+                    offset = None
+                    yield {}
+            elif results.get('query'):
+                answer = results.get('query').get('results')
+                if (answer):
```
danmichaelo: Parenthesis is not needed here. Also, it's better to make it more explicit: `if answer is not None`:

mwclient/client.py
```diff
@@ -979,7 +993,16 @@ def ask(self, query, title=None):
...
-
-            offset = results.get('query-continue-offset')
-            for key, value in results['query']['results'].iteritems():
-                yield {key: value}
+            if results.get('error'):
+                if self.handle_api_result(results):
+                    offset = None
+                    yield {}
+            elif results.get('query'):
+                answer = results.get('query').get('results')
```
danmichaelo: results['query'].get('results')

danmichaelo:
As you noticed, the exception thrown was `KeyError`, not `ApiError`. I wasn't aware that SMW does not use the standard error format, but I will update the `handle_api_result()` method to handle this in another commit once this is merged.

mwclient/client.py
ubibene: In that case answer is an empty list , so 
if answer -> False
if answer is not None -> True
We should keep If answer

ubibene: If the query is valid but there are no results, answer is an empty list,  so 
- if answer -> False
but
- if answer is not None  -> True

We should stick to "if answer"

danmichaelo:
With 9beb051ba6d2feb45cbec5186484959170b920b0 it should now throw `ApiError`

Issue #166, opened 2017-07-26
readapidenied on older mediawiki installations


JohnGreeley:
Login is failing on an older installation of MW (v1.20.2).

An APIError is not being properly handled in mwclient (v0.8.5) which short circuits the attempt to login using the older login mechanism.

PR #168, opened 2017-07-26
Making sure to catch APIError on failed legacy login (#166)

JohnGreeley:
When mwclient attempts to login with an older version of mediawiki, it throws an APIError(readapidenied) from the first failed login, but it fails to catch this typed error and never tries the code with the older MW login mechanism. See #166


danmichaelo:
Thanks! Rebased and merged in f1dc1c2f1030ddb8e54391d7bc71573e871d0c1d

Issue #170, opened 2017-08-11
Ordering of JSON results are not preserved


JohnGreeley:
Currently API results are handled by a vanilla `json.loads`. However, at least with semantic media wiki, you can choose to order results by certain criteria, and json.loads does not preserve the order of these results.

PR #171, opened 2017-08-11
using OrderedDict as object_pairs_hook (#170)

JohnGreeley:
This allows the order of JSON results returned by the api to be preserved. This is important at least for the SMW api, where you can provide ordering criteria in your query. See #170.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -415,7 +415,7 @@ class Site(object):
                             http_method=http_method)
 
         try:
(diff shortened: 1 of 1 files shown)


danmichaelo:
A bit strange that SMW relies on ordering that cannot be guaranteed for by the json spec, but I don't see any disadvantages of using OrderedDict, so I just go ahead and merge this.

PR #172, opened 2017-09-01
README: use SVG version of the logo

waldyrious:
See https://github.com/github/markup/issues/556 for more details.

The rendering is [not entirely correct](https://github.com/github/markup/issues/556#issuecomment-326223588), so don't merge yet, but it [should work soon](https://github.com/github/markup/issues/556#issuecomment-326466252).


waldyrious:
The rendered version preview seems to be working now, so I'll go ahead and merge :D

Issue #156, opened 2017-05-07
Client.ask is returning dictionnary keys only


ubibene:
In client.py :

```
964            for result in results['query']['results']:
965                 yield result 
```

This returns the dictionnary keys of the dictionnary results['query']['results'], but not the dictionnary values .

As such it is not possible to get the answers from the semantic query which are in the dictionnary values.

To get the key and values one may use instead
```
for key, value in results['query']['results'].iteritems():
                yield {key:value}
```

PR #173, opened 2017-10-22
BC: Simplify SMW ask output structure

danmichaelo:
I don't really understand why the `ask()` method yields key-value pairs (`yield {key: value}`) rather than just the values (`yield value`). Since they are yielded from a generator, the keys cannot be used for indexing, so I wonder if this was just a mistake, and I'm tempted to simplify it to yield just the values (result objects), but this will be a breaking change. Do you have any opinions on this, @ubibene ?

As I was querying a site with SMW 2.3.0 I also learned that the format of the results changed either in SMW 2.4.0 or 2.5.0, so I'm adding a check to support both the old format and the new.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -1113,6 +1113,13 @@ class Site(object):
                 query=query, offset=offset), http_method='GET', **kwargs)
             self.handle_api_result(results)  # raises APIError on error
             offset = results.get('query-continue-offset')
-            answers = results['query'].get('results') or {}
-            for key, value in answers.items():
-                yield {key: value}
+            answers = results['query'].get('results', [])
+
+            if isinstance(answers, dict):
+                # In older versions of Semantic MediaWiki (at least until 2.3.0)
+                # a list was returned. In newer versions an object is returned
+                # with the page title as key.
+                answers = [answer for answer in answers.values()]
+
+            for answer in answers:
+                yield answer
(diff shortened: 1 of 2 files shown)


ubibene:
Hi
I'll have a look this week-end.

I remember though that I needed both key and value data. 

I'll keep you posted

ubibene:
Ok , issue  #156 ( https://github.com/mwclient/mwclient/issues/156 ) explains why I needed both keys and values from the generator

danmichaelo:
Not sure if I get you. The problem you desribed in #156 was that it only returned keys, not values. But if we return only values, the key is still part of the value (as `value['fulltext']`), so it should be ok? You could then still get the key:

```python
for result in site.ask('test'):
    key = result['fulltext']
    ...
```

(And looking up by key, something like `site.ask('test')[key]`, never worked anyways, since the result wasn't a dict, but a list of dicts.)

ubibene:
You may be wright Dan.I'm a python newbie and my code is probably far from perfect. Unfortunately as I have changed jobs, I don't have access to "my" code and can't be sure if it will work with minor modifications if you change mwclient the way you suggest.

danmichaelo:
Ok, no prob, just wanted to check with you so I didn't mess up your project!

My guess is that the method is not used much since it was added quite recently to mwclient, but I'll leave this PR open for some time in case someone else want to comment.

PR #177, opened 2017-11-16
Revert "Fix test compat with responses@0.6.0"

AdamWill:
This reverts commit 1650e0e77d370ff5d7b089b60353854d548fb0a1.
Upstream restored handling of the adding_headers arg in 0.6.1.
Going back to using that name means we're still compatible with
any version of responses after 0.3.0, except 0.6.0. Also adjust
setup.py to express this correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,6 @@ setup(name='mwclient',
       install_requires=['requests_oauthlib', 'six'],
       setup_requires=['pytest-runner'],
       tests_require=['pytest', 'pytest-pep8', 'pytest-cache', 'pytest-cov',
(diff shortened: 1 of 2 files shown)


AdamWill:
I'd like this because the package in Fedora is stuck at 0.5.1. I can bump it to a later version for Rawhide, but as the changes in 0.6.0+ are quite big, I couldn't safely send a later version as an update to stable releases without checking everything in the distro that uses responses and making sure it works with that later version. It'd be simpler to have this change in mwclient and then we don't have to ship a responses update for stable releases.

danmichaelo:
Thanks, it's good that we can get back to the latest version!

Issue #179, opened 2017-11-24
 Make pytest-runner a conditional requirement 


yegorich:
Having pytest-runner in setup_requires breaks mwclient build in Buildroot. Could you take similar approach as pytablewriter (https://github.com/thombashi/pytablewriter/commit/63fe493b73437b734e00af68bec1c3bb4407168e#diff-2eeaed663bd0d25b7e608891384b7298)

danmichaelo:
Thanks for the PR! Let me know if you need a new release, otherwise I'll wait with that until another change or two are in.

yegorich:
@danmichaelo thanks for applying the fix. No need to hurry with the release. We've applied the patch locally and will remove it as soon as new release will appear.

PR #180, opened 2017-11-24
setup.py: require pytest-runner only when necessary

bkuhls:
This optimizes setup.py for cases when pytest-runner is not needed,
using the approach that is suggested upstream:

https://pypi.python.org/pypi/pytest-runner#conditional-requirement


bkuhls:
Fixes https://github.com/mwclient/mwclient/issues/179

danmichaelo:
Thanks, this is something I thought about fixing at some point :)

PR #184, opened 2018-01-15
Allow http scheme

alpoza:


danmichaelo:
This is already possible, see https://mwclient.readthedocs.io/en/latest/user/connecting.html

The syntax is a bit awkward, but I've kept it for backwards compability.

PR #198, opened 2018-10-19
Make SSL optional

JohnGreeley:
Adding ssl parameter to allow client to connect to wikis without SSL enabled (useful for internal development sites). SSL is still enabled by default.


danmichaelo:
This patch should not be needed. You can already connect to a non-SSL site. The syntax is a bit weird, but it's documented well:

https://mwclient.readthedocs.io/en/latest/user/connecting.html#connecting-to-your-site

PR #201, opened 2018-11-20
Change invocation syntax to relative Site() in examples

chasemp:
This is what is imported at the beginning of the examples.
This could also be done by changing the import to 'import mwclient'
rather than 'from mwclient import Site'.

Neither seems more correct but I have tried to follow
what seems most intended.

>>> from mwclient import Site
>>> site = mwclient.Site('en.wikipedia.org')
  File "<stdin>", line 1, in <module>
NameError: name 'mwclient' is not defined

vs.

>>> site = Site('en.wikipedia.org')
>>> page = site.pages['Greater guinea pig']
>>> text = page.text()
>>> page.exists
True

### DIFF
--- a/docs/source/user/connecting.rst
+++ b/docs/source/user/connecting.rst
@@ -9,12 +9,12 @@ Begin by importing the Site class:
 
 Then try to connect to a site:
 
-    >>> site = mwclient.Site('test.wikipedia.org')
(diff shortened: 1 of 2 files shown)


danmichaelo:
Good catch, thanks!

PR #204, opened 2018-12-05
Docs: Move reference material into user guide

danmichaelo:
This file was partly outdated and I think it can better be maintained in the user guide. I tried moving a large part of the content into a new `implementation-notes.rst` page, but also incorporated some of it into existing pages.

Opinions? @waldyrious ?


waldyrious:
I agree it makes more sense to have a single source of documentation. Great change! 👍

As for the implementation notes, perhaps we should also merge/move the content currently [in the readme](https://github.com/mwclient/mwclient#implementation-notes)? Or we could keep them there rather than adding a new page. I'm fine with whatever you feel makes the most sense.

danmichaelo:
Ah, yeah, the information in the readme was just the same information. I  removed that, since it's harder to maintain information that is kept in two different places. While looking at the README, I decided to also consolidate the development info. And to make it easier to find, I added a CONTRIBUTING.md, since it's a common place to look for that kind of information.

danmichaelo:
I'm merging this now, but feel free to make any improvements you might see fit!

Issue #109, opened 2016-02-06
Move or copy contents of RELEASE-NOTES.md to the release pages


waldyrious:
Here: https://github.com/mwclient/mwclient/releases. This would prevent duplication.

One relevant question is: would that info stop being tracked in the git repo? Or are the release descriptions kept as (long-form) commit messages in the tags? Even so, would it be preferred to have a text file that could be read in a local clone rather than running a git command to list the tags, or are we ok with providing a link to the github release pages?

@danmichaelo what do you think?

danmichaelo:
I've been thinking about this as well. I don't think the release notes from the Github release pages are available outside Github, and that's been my reason for keeping `RELEASE-NOTES.md` so far, but it's a chore having to update it, and we end up with a lot of "Update release notes" commits, so I would be happy to get rid of it. A link to the github release pages should probably suffice. While Github will go away some day, I don't expect it to happen any time soon.

waldyrious:
Agreed. Current documentation makes sense to live with the repo, but history of changes isn't that critical. I'll take the time to move the notes into the releases page once I get the chance, then.

danmichaelo:
Great!

danmichaelo:
Didn't bother copying everything into RELEASE-NOTES-md yet in case we just remove the file:
https://github.com/mwclient/mwclient/commit/48b7d05f190ac8e0ca7f58497ee17166ce7b4042

waldyrious:
I understand that point of view, and agree that those points are valid concerns. No objection to retain RELEASE-NOTES.md as the primary source of this information, and only copy them to the Github release pages for convenience.

Unfortunately, I can't promise I'll be able to review this and do whatever needs to be done anytime soon, but a checklist of the tasks to perform, according to the points agreed here and on #178, would already be helpful :)

AdamWill:
Well, I was really just registering an opinion - I understand if @danmichaelo doesn't want to go that way, because however you work it it's likely to involve more manual work with each release.

danmichaelo:
I'm ok with keeping the file, but I won't promise that I always remember to update it. We can aim for eventual consistency :)

PR #205, opened 2018-12-09
The changelog

danmichaelo:
In #109 I decided to only update change information in GitHub releases. However, the https://keepachangelog.com/ project have made me rethink that decision and I will try to make the CHANGELOG.md file the primary source for changes from now on.

Ideally, a GitHub release could be generated automatically from the changelog file. I found [gh-release](https://github.com/hypermodules/gh-release), but it's a bit tied to NPM, requiring a package.json file. If someone knows a similar solution that is not tied to NPM, let me know!


waldyrious:
How about [python-semantic-release](https://python-semantic-release.readthedocs.io/en/latest/) ([repo](https://github.com/relekang/python-semantic-release))? It's a Python port of [this npm package](https://www.npmjs.com/package/semantic-release), which provides a way to semi-automatically determine the version [based on the contents of the commit messages](https://github.com/semantic-release/semantic-release#how-does-it-work) included in it.

I think the original package actually [generates the changelog content](https://github.com/semantic-release/release-notes-generator) as well, or at leash has the option to do so. I didn't check if the Python port does the same.

danmichaelo:
I often find that I have to rewrite things a bit when adding them to the changelog. With such an approach, I would have to take some care during the commit process, and probably also clean up commit messages from PRs, but it's definitely something to consider. Perhaps I should try this approach in a personal project just to see if I'm able to follow the strict commit message style at all :)

waldyrious:
Yeah, that was my concern as well. I believe the changelog is generated from PR titles rather than commits, so it shouldn't be hard to edit them after the fact to ensure a mostly clean changelog. But yeah, do try it out first to see if you like it :) such tools should make things easier, not bring added stress to the maintenance process! :P

PR #208, opened 2018-12-22
Improve User-Agent and documentation

### LINKED ISSUE #49: make it easy for users to personalize user-agents (examples, helper method)
Good user-agents for WMF wikis include the script/library name and version, and a way for maintainers to contact the person running the script if needed. 

To make this easy for users, documentation and/or code samples should include examples of individual user/contact information in the user-agent. 

Adding a helper method that makes it easy for logged-out users to build a unique user-agent would also help.

pyfisch:
Use and recommend format from RFC 7231.
Give examples for custom User-Agent strings.

Closes #49

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -28,6 +28,8 @@ __ver__ = '0.9.3'
 
 log = logging.getLogger(__name__)
 
+USER_AGENT = 'mwclient/{} ({})'.format(__ver__, 'https://github.com/mwclient/mwclient')
+
 
 class Site(object):
     """A MediaWiki site identified by its hostname.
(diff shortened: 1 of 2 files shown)


pyfisch:
I think the helper method is not needed. As the preferred grammar is now stated it is easy for consumers to produce correct user agents. If a tool needs more flexibility one can use `format` to create user agent strings.

mwclient/client.py
```diff
@@ -28,6 +28,8 @@
 
 log = logging.getLogger(__name__)
 
+USER_AGENT = 'MwClient/{} ({})'.format(__ver__, 'https://github.com/mwclient/mwclient')
```
pyfisch: Is MwClient or mwclient more correct? Everything except the user agent uses the lower-case variant.

danmichaelo: Good point, We should probably change that to `mwclient`, or, since it's after a period, `Mwclient` I guess.

mwclient/client.py
```diff
@@ -95,14 +97,13 @@ def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
...
-            self.connection.headers['User-Agent'] = (
-                '{prefix}MwClient/{ver} ({url})'.format(
-                    prefix=prefix,
-                    ver=__ver__,
-                    url='https://github.com/mwclient/mwclient'
-                )
-            )
+            # Set User-Agent header field
+            if clients_useragent:
+                ua = clients_useragent + ' ' + USER_AGENT
```
danmichaelo: Would it make sense to change this to `                ua = clients_useragent + '. Using ' + USER_AGENT`?

danmichaelo: Having had a quick glance at RFC 7231, I think the answer to this is "no", so I'm merging this.

PR #213, opened 2019-01-09
Allow URL scheme to be configured in a more intuitive way

### LINKED ISSUE #212: Allow access to site on a http:// URL
Currently, the URL scheme used to access the MediaWiki instance is hardcoded to `https` in the `raw_call` function of `client.py`. This makes it impossible to use mwclient wit a wiki which is hosted under an unencrypted http:// URL.

This should be made configurable (either by adding a scheme parameter to the *Site* constructor or by allowing a full URL instead of just the hostname there).

mtausig:
Add a parameter named `scheme` to the constructor of the `Site` class which allows to specify the URL scheme used. It defaults to *https* (thus not changing the current default behaviour) but it can also be set to *http* to allow access to unencrypted sites.

This fixes #212

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -55,7 +55,7 @@ class Site(object):
                  max_retries=25, wait_callback=lambda *x: None, clients_useragent=None,
                  max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None,
                  reqs=None, consumer_token=None, consumer_secret=None, access_token=None,
-                 access_secret=None, client_certificate=None, custom_headers=None):
+                 access_secret=None, client_certificate=None, custom_headers=None, scheme='https'):
         # Setup member variables
         self.host = host
         self.path = path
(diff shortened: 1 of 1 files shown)


mwclient/client.py
```diff
@@ -364,7 +365,7 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
             headers['Accept-Encoding'] = 'gzip'
         sleeper = self.sleepers.make((script, data))
 
-        scheme = 'https'
+        scheme = self.scheme
         host = self.host
         if isinstance(host, (list, tuple)):
             scheme, host = host
```
waldyrious: Please deprecate this code path, as described [here](https://github.com/mwclient/mwclient/issues/212#issuecomment-453120269) and [here](https://github.com/mwclient/mwclient/issues/212#issuecomment-456792297).

@danmichaelo, would something like [this](https://github.com/mwclient/mwclient/blob/4157e147e07f12219a6b25e976ea87c8d5deff56/mwclient/client.py#L632-L637) be appropriate for the deprecation? Or would we want something supporting Python 3?

danmichaelo: @waldyrious Good idea! I guess DeprecationWarning is still the way to go even though it's hidden by default.

danmichaelo: Done

PR #216, opened 2019-03-08
Make util.parse_timestamp() always return a time.struct_time() object.

marcfrederick:


marcfrederick:
Looking at the tests this seems to be intended.

danmichaelo:
Hi @marcfrederick . Feel free to reopen this. I [added the tests](https://github.com/mwclient/mwclient/commit/8d0650cdd3921066b310f9cbbea33ef1ae274ff5#diff-7678e3487dce054ce8c5c4bdf162e876) this way to maintain backwards compatibility when refactoring the code, but I don't know the original intention behind creating the function this way – and it certainly looks like a wrong way to do things (if it returned `None`, it would be something different).

PR #219, opened 2019-03-09
Make Page.delete() use the "watchlist" parameter

marcfrederick:
Using "watchlist" instead of the deprecated "watch" and "unwatch" parameters. Leaving them as options for backwards compatibility.


waldyrious:
@danmichaelo looks like this one went unnoticed for a while. Maybe we could recover it, what do you think?

PR #220, opened 2019-05-10
Pycodestyle

opoplawski:
pep8 has been replaced with codestyle.

pytest-cache is now included in pytest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,16 @@
 # http://about.travis-ci.org/docs/user/languages/python/
-
-sudo: false
+dist: xenial
 language: python
-
-matrix:
-  include:
-    - python: 2.7
-    - python: 3.4
-    - python: 3.5
-    - python: 3.6
-    - python: 3.7
-      dist: xenial
-      sudo: true
-      # Per https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-426312699
+python:
+  - python: 2.7
+  - python: 3.4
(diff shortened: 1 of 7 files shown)


opoplawski:
Well, that's a bit of a mess then isn't it?  At least for python 3.5+ it seems it's just a matter of dealing with some new checks.  But with 2.7 and 3.4 there seem to be some strange compatibility issues.

danmichaelo:
You are right, we should move from pep8. Annoying thing is that, while `pycodestyle` works with Python 2.7, `pytest-codestyle` does not.

But then there is flake8, which combines pycodestyle, pyflakes and more (mccabe). So my first idea was to try `pytest-flake8` instead, since it works with Python 2.7, but it turned out to have its own quirks (not supporting the same flags as `flake8` itself, for instance `exclude` is [not supported](https://github.com/tholo/pytest-flake8/issues/46#issuecomment-493999666)). It also seems a bit inefficient as it re-initializes flake8 for every single file. And I really dislike that you have to specify flake8 options under the `[pytest]` heading in `setup.cfg`, not under `[flake8]`, which means that you have to duplicate the options if you also would like to support running flake8 standalone.

So... wait… do we really need to do linting and style checking through pytest? Probably not :) It seems like we will get a more clean and stable setup by letting pytest just run the tests, and then run flake8 (or pycodestyle or whatever) separately, so I will prepare that.

PR #223, opened 2019-06-26
Remove one more stray reference to pep8

AdamWill:
We switched to pycodestyle then flake8 recently, but one stray
reference to pep8 was left in this doc file.

### DIFF
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -46,7 +46,7 @@ and run pytest directly:
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


AdamWill:
note, I think this line should also have `pytest-cov` in it, but I'm not 100% sure.

danmichaelo:
Thanks! Yes, indeed, pytest-cov is also needed when running `py.test`. I added that now

danmichaelo:
Great branch name btw. 😄

danmichaelo:
And `mock`

Issue #224, opened 2019-06-26
Tox doesn't actually run any tests any more


AdamWill:
57f716ed removed the `commands` line from `tox.ini`, this one:

    commands=py.test -v --cov --codestyle mwclient test

I'm not sure if this was a mistake or somehow intentional, but the result is that calling `tox` doesn't run the test suite any more. It *only* runs flake8.

danmichaelo:
Definitely a mistake, sorry about that! Thanks for the fix!

PR #225, opened 2019-06-26
tox.ini: restore commands line, fix tabs vs. spaces (#224)

AdamWill:
The `commands` line was accidentally(?) removed entirely in
57f716e, so tox no longer actually ran any tests (only flake8).
This restores it, with the codestyle bit removed obviously. It
also fixes two lines where a tab was used instead of spaces.


danmichaelo:
Thanks!

PR #227, opened 2019-06-27
Update dependency to work on conda

JohnGreeley:
requests_oauthlib dependency name will only install from pip, requests-oauthlib works on both conda and pip


danmichaelo:
Most interesting 😮 Thanks!

PR #131, opened 2016-07-24
Refactor

lukasjuhrich:
__Please don't merge yet!__

These are random improvements I made without a lot of order.

I want to start a discussion which changes are good, and which are disliked (probably the line wrappings). Please comment on the according file changes below.

I tried to add docstrings where things were unclear for me or use a more „pythonic“ way.

I hope the docstring format is correct, since I do not know anything about sphinx or similiar.

I probably will continue to push commits onto this pull request.

PR #228, opened 2019-08-03
Expanding the .editorconfig file

marcfrederick:
- Set the top level `.editorconfig` file as the root (good practice).

A bunch of rules to match the existing conditions:
- Indent `.cfg`, as well as `Makefile`s using tabs instead of spaces.
- Set an indent-size of 2 instead of 4 for `.yml` and `.yaml` files.
- Set a max line length of **90** (non standard) for Python files, as this seems to best match the existing code-base, though some changes will be necessary to make it fully compliant. Alternatively I'd like to suggest a max line length of either 79 (pep8) or just going up to something like 100 or even 120 characters.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,7 +16,7 @@ indent_size = tab
 indent_style = tab
 
 [*.py]
-max_line_length = 88
+max_line_length = 90
 
 [*.{yaml, yml}]
 indent_size = 2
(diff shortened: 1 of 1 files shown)


danmichaelo:
Much better!

You're right that the code base matches a max line length of 90 characters quite well. This was actually standardized in #131, but then got lost at some point :/ I did a small reformat now (1e7e496aa6326f67b16e4444a583c22aedb61543), so we can bring the limit back.

PR #230, opened 2019-08-04
Improve README

waldyrious:
- Center title to match the logo
- Change logo image to the PNG version, as SVG seems to be unsupported by PyPI:
  https://pypi.org/project/mwclient/
- Remove Landscape.io badge, as the service doesn't seem to be working:
  - https://landscape.io/github/mwclient/mwclient
  - https://github.com/landscapeio/landscape-issues/issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
-<div align="center"><img src="docs/source/logo.svg" width="300"/></div>
-
-# mwclient
+<div align="center">
+    <img src="docs/source/logo.png"/>
+    <h1>mwclient</h1>
+</div>
 
 [![Build status][build-status-img]](https://travis-ci.org/mwclient/mwclient)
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,12 @@
-<div align="center"><img src="docs/source/logo.svg" width="300"/></div>
-
-# mwclient
+<div align="center">
+    <img src="docs/source/logo.png"/>
```
danmichaelo: While we're at it, should we also add an `alt` attribute here?

danmichaelo:
Thanks, yes, the Landscape service has always been quite unstable. It's a nice service when it works though, so I hope it will eventually recover and made a small donation to support the development, but I agree we can remove the the badges now, and rather add them back later if the service recovers.

README.md
waldyrious: Good call, will do.

waldyrious: Done.

Issue #51, opened 2014-07-08
Confusing function names


fhocutt:
It is confusing that the user calls <code>page.edit()</code> to retrieve the text of a page, and <code>page.save(...)</code> to edit the text of a page. Consider deprecating/renaming these functions to improve usability.

PR #232, opened 2019-08-27
Add append/prepend setting in save & use this to add touch

RheingoldRiver:
Add support for using appendtext or prependtext instead of text in an api edit request. This also allows a graceful way to blank/null edit a page without having to spend two api queries to get the text and then save, so add page.touch() as well using this.

### DIFF
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -171,10 +171,28 @@ class Page(object):
             self._textcache[key] = text
         return text
 
-    def save(self, text, summary=u'', minor=False, bot=True, section=None,
-             edit_type='replace', **kwargs):
+    def save(self, *args, **kwargs):
+        """Alias for edit, for maintaining backwards compatibility."""
+        return self.edit(*args, **kwargs)
+
+    def edit(self, text, summary=u'', minor=False, bot=True, section=None, **kwargs):
         """Update the text of a section or the whole page by performing an edit operation.
         """
+        return self._edit(summary, minor, bot, section, text=text, **kwargs)
+
+    def append(self, text, summary=u'', minor=False, bot=True, section=None,
+                    **kwargs):
+        """Append text to the text of a section or the whole page by performing an edit operation.
+        """
+        return self._edit(summary, minor, bot, section, appendtext=text, **kwargs)
+
+    def prepend(self, text, summary=u'', minor=False, bot=True, section=None,
+                     **kwargs):
+        """Prepend text to the text of a section or the whole page by performing an edit operation.
+        """
+        return self._edit(summary, minor, bot, section, prependtext=text, **kwargs)
+
+    def _edit(self, summary, minor, bot, section, **kwargs):
         if not self.site.logged_in and self.site.force_login:
             raise mwclient.errors.AssertUserFailedError()
         if self.site.blocked:
@@ -205,21 +223,9 @@ class Page(object):
             data['assert'] = 'user'
 
         def do_edit():
-            result = None
-            if edit_type == 'replace':
-                result = self.site.post('edit', title=self.name, text=text,
-                                        summary=summary, token=self.get_token('edit'),
+            result = self.site.post('edit', title=self.name, summary=summary,
+                                    token=self.get_token('edit'),
                                         **data)
-            elif edit_type == 'append':
-                result = self.site.post('edit', title=self.name, appendtext=text,
-                                        summary=summary, token=self.get_token('edit'),
-                                        **data)
-            elif edit_type == 'prepend':
-                result = self.site.post('edit', title=self.name, prependtext=text,
-                                        summary=summary, token=self.get_token('edit'),
-                                        **data)
-            else:
-                raise mwclient.errors.InvalidParameter('edit_type', edit_type)
             if result['edit'].get('result').lower() == 'failure':
                 raise mwclient.errors.EditError(self, result['edit'])
             return result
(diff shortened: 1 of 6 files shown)


mwclient/page.py
```diff
@@ -204,9 +205,21 @@ def save(self, text, summary=u'', minor=False, bot=True, section=None, **kwargs)
             data['assert'] = 'user'
 
         def do_edit():
-            result = self.site.post('edit', title=self.name, text=text,
-                                    summary=summary, token=self.get_token('edit'),
-                                    **data)
+            result = None
+            if edit_type == 'replace':
+                result = self.site.post('edit', title=self.name, text=text,
```
danmichaelo: Hm, it's not so elegant to repeat this code three times. I wonder if it would be better to add two new methods `append(text)` and `prepend(text)` that calls `save(...)`

RheingoldRiver: How about this:
```
    def save(self, text, summary=u'', minor=False, bot=True, section=None, **kwargs):
        """Update the text of a section or the whole page by performing an edit operation.
                """
        return self._save(summary, minor, bot, section, text=text, **kwargs)
    
    def save_append(self, text, summary=u'', minor=False, bot=True, section=None,
                    **kwargs):
        """Append text to the text of a section or the whole page by performing an edit operation.
                """
        return self._save(summary, minor, bot, section, append_text=text, **kwargs)
    
    def save_prepend(self, text, summary=u'', minor=False, bot=True, section=None,
                     **kwargs):
        """Prepend text to the text of a section or the whole page by performing an edit operation.
                """
        return self._save(summary, minor, bot, section, prepend_text=text, **kwargs)
```

and then below in _save,
```
 def do_edit():
            result = self.site.post('edit', title=self.name, summary=summary,
                                    token=self.get_token('edit'),
                                        **data)
```

I didn't test yet but if that looks good to you I'll run tests, fix the docstring for touch, and send a new commit - sorry for the whitespace issues, I'll try to fix those too.

danmichaelo: Did you see my other suggestion above? I think I would prefer having methods named just `append` and `prepend` rather  than `save_prepend` and `save_append`. Do you agree?

RheingoldRiver: Prepend/Append didn't sound obviously enough like page-saving actions to me out of context, and if they start with save_ then they'd show up in autocomplete along with save, is why I renamed them. But if you'd still prefer just prepend/append I'm fine with that just to have simpler names for everything.

danmichaelo: I still find the `save_` names a bit clumsy, but it sounds useful to have them show up in autocomplete along with save, yes, I agree. We could try getting a third opinion on this. @waldyrious, do you have a preference here?

waldyrious: I understand the motivation, but I am not sure the `save_` prefix is the best solution to the problem, for a few  reasons:

1. We already have other page-changing methods that aren't prefixed with `save_`, such as `delete()` and `move()`
2. "save" is actually a little ambiguous, since it doesn't imply that the text is changed. It sounds more like a method that's part of a framework that supported piling up local changes and then a `save()` method to commit them.
3. The names `save_prepend` and `save_append` are, indeed, a little odd-sounding IMO. They don't feel like part of a well-planned API.

My suggestion would be to follow-up #51 and rename `save()` to `edit()` (which should be reasonable, since it's been three releases since the one that changed `edit()` to `text()`), and keep the two new methods named `prepend()` and `append()`.

How does that sound?

RheingoldRiver: Would you keep save() as an alias to edit() for backwards compatibility? if so that sounds good, otherwise I'm not sure it's worth breaking compatibility to rename save, save -> edit is only a small improvement in clarity compared to edit -> text.

waldyrious: Sure, no objections to keeping `save()` as an alias for compatibility.

RheingoldRiver: Ok great, I have a pretty bad cold right now but I'll send new changes by the end of the week:
* edit, prepend, and append will call _edit
* _edit will be similar to current save except text/append_text/prepend_text will be part of **kwargs
* alias of save for edit
* update touch to work with new method & add clarity in docstring

danmichaelo: Wish you a quick recovery  :sparkles:

Sounds like a good plan. My first thought was that `append` and `prepend` could just call <s>`save`</s> `edit(text=None, ...)`, but I'm open to the idea of adding a new `_edit` method if you think that's less messy :)

RheingoldRiver: Ok made the changes. Do you have a settings file I can import to Pycharm so I have the right whitespace settings for everything?

waldyrious: > Do you have a settings file I can import to Pycharm so I have the right whitespace settings for everything?

`.editorconfig` files are [natively supported by PyCharm](https://editorconfig.org/#download), so it should have picked up the definitions [we have already](https://github.com/mwclient/mwclient/blob/master/.editorconfig). Maybe there are additional configurations we need to add to the file?

RheingoldRiver:
> perhaps we should just keep it indefinitely as an alias to avoid breaking things? I think I'm lending towards that.

Agreed, I can't imagine wanting to use that name for something else, and just thinking about my own mwclient usage, that's so very many things I'd have to change.

Updated the tests and docs, and fixed the line that was 91 characters long.

waldyrious:
Thanks for the contribution and your patience, @RheingoldRiver! 🎉

PR #234, opened 2019-09-24
doc: Update index.rst

MarcoFalke:
This fixes the warning:

```
DeprecationWarning: Specifying host as a tuple is deprecated as of mwclient 0.10.0. Please use the new scheme argument instead.


danmichaelo:
Thanks! I guess this should be fixed more places too

MarcoFalke:
Indeed. Found one more instance.

PR #235, opened 2019-09-30
Update LICENSE.md

waldyrious:
- Update copyright notice to reflect changes in recent years
- Add title to make the license file more human-friendly

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 ## MIT License
 
-Copyright (c) 2006-2019 Bryan Tong Minh, Dan Michael O. Heggø and contributors
+Copyright (c) Bryan Tong Minh, Dan Michael O. Heggø and contributors
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,6 @@
-Copyright (c) 2006-2013 Bryan Tong Minh
+## MIT License
+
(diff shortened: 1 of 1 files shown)


danmichaelo:
Thanks! What do you think about removing the year range altogether? Like here: https://github.com/facebook/react/blob/master/LICENSE

Or alternatively just use the date of first publication (2006) rather than a date range?

waldyrious:
Both approaches are OK with me. There are [compelling arguments](https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/) to use a date range, but the no-date approach of React seems like a nice shortcut to signify "all time" (and avoid the need to constantly update the year range).

If after reading the page linked above you still think that using no date is reasonable, I'll make the change.

danmichaelo:
Thanks, that's a well-written post, but I'm a bit confused about the arguments, it seems to boil down to precedence, doesn't it?

Sticking to precedence or standards can be a reason by itself though. But having a date range that becomes outdated every year (unless the project development actually stops, that is) still seems a bit clumsy to me when the whole copyright statement doesn't really carry any legal function anymore. 

It still seems reasonable to me to remove the date range, but I'm also OK with keeping it. At least if you could help me update it at least every third year or so 😄

waldyrious:
IMO if the project changed licenses (or license holders), a date range would be relevant. As it stands, it doesn't really matter that much to have the boundaries there, since no project with the same name or content (and different licensing conditions) exists outside them anyway. Let's go with the easier solution then ;)

danmichaelo:
> IMO if the project changed licenses (or license holders), a date range would be relevant

That's true!

PR #238, opened 2019-11-13
add methods to manage users:

DylannCordel:
* simple user creation
* block user
* unblock user
* get basic info about user
* get groups
* add groups
* remove groups
* set groups (get current ones and add/remove what is needed)


mwclient/errors.py
```diff
@@ -14,6 +14,10 @@ class MaximumRetriesExceeded(MwClientError):
     pass
 
 
+class MwClientNotFound(MwClientError):
```
danmichaelo: Hm, this name implies that it's more generic than it currently is. Do you see other use cases for it? If not, I think it's better to be specific and name it `UserNotFound`. Especially since the other exceptions are generally quite specific (and do not use the Mw prefix).

mwclient/client.py
```diff
@@ -50,6 +50,9 @@ class Site(object):
         >>> site = mwclient.Site('sourceforge.net', path='/apps/mediawiki/mwclient/')
 
     """
+    AVAILABLE_TOKENS_TYPES = {
```
danmichaelo: TOKEN_TYPES, not TOKENS_TYPES perhaps ? And while we're touching this, perhaps the list should also be sorted alphabetically for better readability?

mwclient/client.py
```diff
@@ -612,6 +615,156 @@ def get_token(self, type, force=False, title=None):
 
         return self.tokens[type]
 
+    def create_user(self, username, password, **kwargs):
+        if 'createtoken' not in kwargs:
+            kwargs['createtoken'] = self.get_token('createaccount')
+        if 'retype' not in kwargs:
+            kwargs['retype'] = password
+        if 'continue' not in kwargs and 'createreturnurl' not in kwargs:
+            # FIXME should be great if API didn't require this...
```
danmichaelo: The comment is OK, but please remove "FIXME" if it cannot be fixed without a change in the API.

mwclient/client.py
```diff
@@ -612,6 +615,156 @@ def get_token(self, type, force=False, title=None):
...
+            kwargs['createreturnurl'] = '%s://%s' % (self.scheme, self.host)
+        info = self.post('createaccount', username=username, password=password, **kwargs)
+        if info['createaccount']['status'] == 'FAIL':
+            raise Exception('Unable to createaccount [%s]: %s' % (
```
danmichaelo: Please use a subclass instead. APIError perhaps?

mwclient/client.py
```diff
@@ -612,6 +615,156 @@ def get_token(self, type, force=False, title=None):
...
+                info['createaccount']['messagecode'], info['createaccount']['message'])
+            )
+        return True
+
+    def get_user(
+        self, username=None, userid=None,
+        usprop='registration|groups|blockinfo'
+    ):
+        if (not username and not userid) or (username and userid):
+            raise Exception('username OR userid are required')
```
danmichaelo: Please use a subclass instead. [ValueError](https://stackoverflow.com/questions/256222/which-exception-should-i-raise-on-bad-illegal-argument-combinations-in-python) perhaps?

mwclient/client.py
```diff
@@ -612,6 +615,156 @@ def get_token(self, type, force=False, title=None):
...
+
+    def unblock_user(self, username=None, userid=None, reason=None, tags=None):
+        return self._block_unblock_user(False, username, userid, reason, tags)
+
+    def _block_unblock_user(
+        self, block=True, username=None, userid=None,
+        reason=None, tags=None
+    ):
+        if (not username and not userid) or (username and userid):
+            raise Exception('username OR userid are required')
```
danmichaelo: ValueError?

mwclient/client.py
```diff
@@ -612,6 +615,156 @@ def get_token(self, type, force=False, title=None):
...
+            kwargs['userid'] = userid
+        if reason:
+            kwargs['reason'] = reason
+        if tags:
+            kwargs['tags'] = '|'.join(tags)
+        return self.post('block' if block else 'unblock', **kwargs)
+
+    def get_user_groups(self, username=None, userid=None):
+        if (not username and not userid) or (username and userid):
+            raise Exception('username OR userid are required')
```
danmichaelo: ValueError?

DylannCordel:
Thanks for the review @danmichaelo 

Requested changes are done and I also enhance coverage of new lines of code.

mwclient/errors.py
DylannCordel: I renamed `MwClientNotFound` to `UserNotFound` and creates new ones to get
  better debugging possibilities:
    * `NotFound`
    * `CreateError`
    * `UserCreateError`

mwclient/client.py
DylannCordel: done

mwclient/client.py
DylannCordel: done

mwclient/client.py
DylannCordel: done

mwclient/client.py
DylannCordel: done

mwclient/client.py
DylannCordel: done

mwclient/client.py
DylannCordel: done

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+        if 'retype' not in kwargs:
+            kwargs['retype'] = password
+        if 'continue' not in kwargs and 'createreturnurl' not in kwargs:
+            # should be great if API didn't require this...
+            kwargs['createreturnurl'] = '%s://%s' % (self.scheme, self.host)
+        info = self.post('createaccount', username=username, password=password, **kwargs)
+        if info['createaccount']['status'] == 'FAIL':
+            raise errors.UserCreateError(
+                info['createaccount'].get('messagecode', 'unknown'),
+                info['createaccount'].get('message', 'No reason...'),
```
danmichaelo: The defaults are a bit arbritrary. I suggest removing 'unknown' and 'No reason...' and just leave the default null value (`None`).

mwclient/errors.py
```diff
@@ -22,6 +22,36 @@ def __init__(self, code, info, kwargs):
...
+        super(NotFound, self).__init__(code, info, kwargs)
+
+
+class CreateError(APIError):
+
+    def __init__(self, entity_type, code=None, info=None, kwargs=None):
+        if not info:
+            info = 'Entity "%s" can not been created.' % entity_type
+        self.entity_type = entity_type
+        super(CreateError, self).__init__(code, info, kwargs)
```
danmichaelo: I'm a bit uneasy about `CreateError` and `NotFound` as I don't really see the use for them. It's useful to be able to distinguish between *a specific* API error response (`UserCreateError`) and *any* API error response (`APIError`), but I don't see the need for one more level between those. Even if we create a new `PageCreateError` and makes it extend `CreateError`, I don't see the need to catch *some kind of creation failure* (either user or page).

Could you elaborate why you think these can be useful if you want to keep them?

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+             ...                                groups=['sysop', 'bureaucrat'])
+             ... except mwclient.errors.APIError:
+             ...     print('Can not add groups to user: %s' % e)
+             ... else:
+             ...     print('really added groups: %s ; really removed groups: %s',
+             ...            res['added'], res['removed'])
+
+         Args:
+             username (str): User name to unblock, OR
+             userid (int): User ID to unblock
```
danmichaelo: unblock seems like a copy-paste error ;)

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Removes groups to the user
+
+             >>> try:
+             ...     site.remove_user_groups('SomeUser', groups=['sysop', 'bureaucrat'])
+             ... except mwclient.errors.APIError:
+             ...     print('Can not remove groups to user: %s' % e)
+
+         Args:
+             username (str): User name to unblock, OR
+             userid (int): User ID to unblock
```
danmichaelo: unblock seems like a copy-paste error

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Adds groups to the user
+
+             >>> try:
+             ...     site.add_user_groups('SomeUser', groups=['sysop', 'bureaucrat'])
+             ... except mwclient.errors.APIError:
+             ...     print('Can not add groups to user: %s' % e)
+
+         Args:
+             username (str): User name to unblock, OR
+             userid (int): User ID to unblock
```
danmichaelo: unblock seems like a copy-paste error

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Retrieves groups the user belongs to
+
+             >>> try:
+             ...     site.get_user_groups('SomeUser')
+             ... except mwclient.errors.APIError:
+             ...     print('Can not retrieves user\'s groups: %s' % e)
+
+         Args:
+             username (str): User name to unblock, OR
+             userid (int): User ID to unblock
```
danmichaelo: unblock seems like a copy-paste error

mwclient/errors.py
DylannCordel: Mmm yes you right. Seemed to be a good idea when I added this to have a common Exception, allowing contributors to add future *NotFound subclasses.

mwclient/client.py
DylannCordel: done

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+             username (str): User name to retrieve, OR
+             userid (int): User ID to retrieve
+
+         Returns:
+             Dictionary of the JSON user informations response
+
+         Raises:
+             ValueError: username and userid params are both not set or set
+             UserNotFound (mwclient.errors.UserNotFound): User has not been found
+             APIError (mwclient.errors.APIError): Other API errors
```
danmichaelo, line 687: Indentation should be changed from 9 to 8 spaces. Please check the other docstrings too, it seems like most are indented 9 characters.

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+        """
+        if (not username and not userid) or (username and userid):
```
danmichaelo, line 689: I would suggest

    if (username is None) == (userid is None):

as it's slightly more compact, but this is a matter of taste, I guess

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+            raise ValueError('username OR userid are required')
+
+        kwargs = {
+            'list': 'users',
+            'usprop': usprop,
+        }
+        if username:
+            kwargs['ususers'] = username
+        elif userid:
```
danmichaelo, line 698:
```suggestion
        else:
```

 since userid cannot be None given the test above.

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+             APIError (mwclient.errors.APIError): Other API errors
+        """
+        if (not username and not userid) or (username and userid):
+            raise ValueError('username OR userid are required')
+
+        kwargs = {
+            'list': 'users',
+            'usprop': usprop,
+        }
+        if username:
```
danmichaelo, line 696:
```suggestion
        if username is not None:
```

It's generally safer to be more explicit. https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+            kwargs['ususers'] = username
+        elif userid:
+            kwargs['ususerids'] = userid
+
+        resp = self.get('query', **kwargs)
+
+        if 'missing' in resp['query']['users'][0]:
+            raise errors.UserNotFound(code='missing', info=None, kwargs=kwargs)
+        return resp['query']['users'][0].get('groups', [])
```
danmichaelo, line 820:
```suggestion
        user = self.get_user(username, userid, 'groups')

        return user.get('groups', [])
```

to reduce the amount of code duplication.

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+             ValueError: username and userid params are both not set or set
+             UserNotFound (mwclient.errors.UserNotFound): User has not been found
+             APIError (mwclient.errors.APIError): Other API errors
+        """
+        if (not username and not userid) or (username and userid):
+            raise ValueError('username OR userid are required')
+
+        kwargs = {
+            'list': 'users',
+            'usprop': usprop,
```
danmichaelo, line 694:
```suggestion
            'usprop': prop,
```

per above

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+        if username:
+            kwargs['ususers'] = username
+        elif userid:
+            kwargs['ususerids'] = userid
+        resp = self.get('query', **kwargs)
+        if 'missing' in resp['query']['users'][0]:
+            raise errors.UserNotFound(code='missing', info=None, kwargs=kwargs)
+        return resp['query']['users'][0]
+
+    def block_user(self, username=None, userid=None, reason=None, tags=None):
```
danmichaelo, line 705:
```suggestion
    def block_user(self, username=None, userid=None, reason=None, tags=None, **kwargs):
```

since there are many more possible arguments too

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Blocks an user
+
+             >>> try:
+             ...     site.block_user('SomeUser')
+             ... except mwclient.errors.APIError:
+             ...     print('Can not block the user: %s' % e)
+
+         Args:
+             username (str): User name to block, OR
+             userid (int): User ID to block
```
danmichaelo, line 716: Add

```
             reason (str): Block reason
             tags (list): List of tags
             **kwargs (dict): Additional arguments are passed on to the API
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Returns:
+             Dictionary of the JSON block informations response
+             (see https://www.mediawiki.org/wiki/API:Block)
+
+         Raises:
+             ValueError: username and userid params are both not set or set
+             APIError (mwclient.errors.APIError): API errors (see codes available
+                       on https://www.mediawiki.org/wiki/API:Block#Possible_errors)
+        """
+        return self._block_unblock_user(True, username, userid, reason, tags)
```
danmichaelo, line 727:
```suggestion
        return self._block_unblock_user(True, username, userid, reason, tags, **kwargs)
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+
+    def unblock_user(self, username=None, userid=None, reason=None, tags=None):
```
danmichaelo, line 729:
```suggestion
    def unblock_user(self, username=None, userid=None, reason=None, tags=None, **kwargs):
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+         Returns:
+             Dictionary of the JSON unblock informations response
+             (see https://www.mediawiki.org/wiki/API:Block)
+
+         Raises:
+             ValueError: username and userid params are both not set or set
+             APIError (mwclient.errors.APIError): API errors (see codes available
+                       on https://www.mediawiki.org/wiki/API:Block#Possible_errors)
+        """
+        return self._block_unblock_user(False, username, userid, reason, tags)
```
danmichaelo, line 751:
```suggestion
        return self._block_unblock_user(False, username, userid, reason, tags, **kwargs)
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+
+    def _block_unblock_user(
+        self, block=True, username=None, userid=None,
+        reason=None, tags=None
```
danmichaelo, line 755:
```suggestion
        reason=None, tags=None, **kwargs
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+    ):
+        """
+        Blocks or unblocks a user. You should not use this protected method: prefers
+        the use of block_user or unblock_user. See docs of theses public methods.
+        """
+        if (not username and not userid) or (username and userid):
+            raise ValueError('username OR userid are required')
+        kwargs = {
+            'token': self.get_token('csrf'),
+        }
```
danmichaelo, line 765:
```suggestion
        kwargs['token'] = self.get_token('csrf')
```

mwclient/client.py
```diff
@@ -612,6 +616,380 @@ def get_token(self, type, force=False, title=None):
...
+        if username:
+            kwargs['user'] = username
+        if userid:
```
danmichaelo, line 768:
```suggestion
        else:
```

mwclient/client.py
waldyrious: To be honest, you suggestion is harder to reason about. I would prefer @DylannCordel's approach as it's more readable.

danmichaelo: Thanks, I can see that it can be, so let's keep the suggestion then :)

DylannCordel:
Hi @danmichaelo 

Requested changes have been applied. I have not the "Allow edits from maintainers" link on this PR. Maybe it's only when we create a new-one ?

waldyrious:
> I have not the "Allow edits from maintainers" link on this PR. Maybe it's only when we create a new-one ?

You should be able to see it as a checkbox in the sidebar of this PR, as shown in the screenshot from the page @danmichaelo linked to:

"Allow edits from maintainers" checkbox screenshot

Don't you see anything there?

DylannCordel:
> You should be able to see it as a checkbox in the sidebar of this PR, as shown in the screenshot from the page @danmichaelo linked to:

Not in this PR. On other projects, I have this checkbox (eg: https://github.com/IdentityPython/djangosaml2/pull/292) but on this one I have not the checkbox. Maybe it's because the repo is inside our organization namespace instead my personal namespace. I looked into organization settings and repository settings to find an option to allow it but nothing seems to do the job. It seems to be an old known problem : https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847

I'll fork into my personal account next time.

waldyrious:
> Maybe it's because the repo is inside our organization namespace instead my personal namespace.

Ah, that makes sense. Yeah, it should be it. Thanks for digging up that info!

DylannCordel:
move to #299 to allow edits by maintainers

PR #239, opened 2019-11-20
Update regarding mock

havardAasen:
mock was included in unittest in python 3.3 unittest.mock
It is backported in PyPi so it should work either way up to python 3.6, altough i prefer the built-in.

The test files was tested with python 3.7 and 3.8 and the try, except worked. The setup.py file is not tested.


danmichaelo:
Merged in 8d6e48533438fc669c0d9672dc85101b2960c737

PR #241, opened 2019-11-21
transform MW deprecation warnings into python DeprecationWarning

DylannCordel:
Try to avoid fake warnings which are just deprecation warnings  to keep logs cleans and usefull

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -307,8 +307,21 @@ class Site(object):
         self.logged_in = 'anon' not in userinfo
         if 'warnings' in info:
             for module, warning in info['warnings'].items():
-                if '*' in warning:
-                    log.warning(warning['*'])
+                msg = warning.get('*')
+                if not msg:
+                    continue
+                if (
+                    'is deprecated' in msg
+                    or (
+                        module == 'main'
+                        and 'Subscribe to the mediawiki-api-announce mailing' in msg
+                    )
+                ):
+                    # try to avoid fake warnings which are just deprecation warnings
+                    # to keep logs cleans and usefull
+                    warnings.warn(msg, DeprecationWarning)
+                else:
+                    log.warning(msg)
 
         if 'error' in info:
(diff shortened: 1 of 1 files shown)


danmichaelo:
Sorry about the very late response! As for the `None` check, best practice is to use "is not None":

https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html#compare-values-to-none-using-the-pattern-if-cond-is-none

Overall, I'm a little bit reluctant about this PR. The deprecation warnings from the API are really warnings in the true sense, as they warn you about something you need to address to avoid the code from breaking in the future. Esp. for scripts that rarely needs to be touched, a sudden rise of warnings can be a good reminder about the existence of the script, and that something needs to be done with it at some point :)

Warnings from the `warnings` module are often easier to miss. But that can be addressed with `logging.captureWarnings`:

https://docs.python.org/2/library/logging.html#logging.captureWarnings

So perhaps a compromise could be to merge this PR, but mention in the documentation that `logging.captureWarnings(true)` can be used to log deprecation warnings from the API? Any thoughts on this, @waldyrious ?

In any case, I would prefer to remove the wording "fake warnings" ;)

waldyrious:
> Any thoughts on this, @waldyrious ?

I never really user mwclient as a library in another project (I only made quick scripts here and there, back when I dabbled a little with Python), so I don't think I'm qualified to comment on the signal-to-noise ratio of deprecation warnings.

On principle, I do agree with what you say above:

> The deprecation warnings from the API are really warnings in the true sense, as they warn you about something you need to address to avoid the code from breaking in the future.

So my inclination, for what it's worth, would be to have deprecation warnings shown by default in the output.

DylannCordel:
Hi guys,

Some news about this PR ? 

This commit does not disable the "deprecated messages" from the logs but transforms those into a real `DeprecationWarning`. Those ones can be displayed or not in the console depending the logging configuration.

AdamWill:
I don't have a really strong opinion on it either way, personally...

PR #243, opened 2019-12-11
remove: py34 references

marcfrederick:
Support for Python 3.4 is stated in multiple places and already tested by the tox.ini

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 mwclient is a lightweight Python client library to the
 [MediaWiki API](https://mediawiki.org/wiki/API)
 which provides access to most API functionality.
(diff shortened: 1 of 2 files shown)


waldyrious:
This had been added before, in 22ac6ecb8. It was later removed in 737e7cc26.

Can you point to the places where support for Python 3.4 is claimed? We probably should remove them.

marcfrederick:
@waldyrious I've updated this pull request to remove the references to Python 3.4 instead.

waldyrious:
Thanks, @marcfrederick! Can you just fix the commit message? It currently says "remove: py37 references" but it should be "remove: py34 references" instead.

marcfrederick:
Done @waldyrious

PR #244, opened 2019-12-12
remove: simplejson dependency

marcfrederick:
This library currently uses `simplejson` instead of `json` conditionally in multiple places. Presumably, this was done for compatibility with Python versions prior to 2.6, when `json` was added to the standard library. As this library only supports 2.7+, this should not be needed anymore.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -6,10 +6,7 @@ import six
 
 from collections import OrderedDict
 
-try:
-    import json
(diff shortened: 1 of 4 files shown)


danmichaelo:
Yes, the fallback to simplejson was indeed needed for Python 2.6, so it can safely be removed now. Thanks! (And sorry for the slow response)

PR #245, opened 2020-02-02
Add base_title & base_name attributes to page

RheingoldRiver:
I find these two attributes to be pretty useful to have easily available

### DIFF
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -45,6 +45,9 @@ class Page(object):
         else:
             self.page_title = self.name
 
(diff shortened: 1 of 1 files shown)


RheingoldRiver:
Hi, just wanted to bump this - would you be ok with this change?

danmichaelo:
Sorry for not responding earlier! Yes, I can see that those additions can be useful, so I'll go ahead and merge this.

On a side note, the naming of `page.name` and `page.page_title` have always annoyed me, but it would break too many things to change, and you kinda get used to it.

PR #247, opened 2020-02-11
remove: ex.py

marcfrederick:
Seems to be a leftover. None of the functions are used, tested, or documented. Most aren't even compatible with py3.


danmichaelo:
Thanks! I've been on the verge of deleting this myself.

PR #253, opened 2020-05-30
remove custom null handler

marcfrederick:
Not required as `logging.NullHandler` is available on all supported python versions. See [here](https://docs.python.org/2/library/logging.handlers.html#nullhandler) and [here](https://docs.python.org/3/library/logging.handlers.html#nullhandler).


danmichaelo:
Ah, another remain from Python 2.6. Thanks!

PR #254, opened 2020-06-08
Retry api calls on timeout

trueg:
While I should probably manage to solve the timeout problems I experience every once in a while, I also think it would make sense to retry.


danmichaelo:
Yes, I agree. Will leave the PR open for a few days though, just in case anyone else have comments.

PR #264, opened 2020-11-22
Add move_subpages, ignore_warnings to page.move

RheingoldRiver:


AdamWill:
Looks fine to me, these options are documented at https://wiki.creativecommons.org/api.php?action=help&modules=move and seem to have been in the code forever AFAICT (see https://github.com/wikimedia/mediawiki/blob/master/includes/api/ApiMove.php ) so we don't need to worry much about them not being available on older mediawiki or anything.

waldyrious:
Hey @RheingoldRiver, thanks so much for the outsized patience with how long it took for this to get merged.

As a recurring contributor, would you be interested in helping maintain this project? Please see https://github.com/mwclient/mwclient/issues/280 and let us know!

PR #269, opened 2021-09-26
setup.py: update url

lilydjwg:
I don't know what happened, but this url definitely needs an update :-)


waldyrious:
Good call, thanks! :)

gene1wood:
As there hasn't been a release since this fix, the old URL is still live on PyPi : https://pypi.org/project/mwclient/

You may want to increment the version, cut a release and upload it to PyPi to fix the URL

waldyrious:
Unfortunately, as I'm not a maintainer of the package in PyPi, I cannot do that (I don't know how to do it either, for that matter). We'll have to wait for @danmichaelo to complete those steps.

PR #273, opened 2022-02-09
Add mime to imageinfo

mitar:
Very useful when dealing with images.


AdamWill:
Looks good, I tested this locally and it does what's expected, the mime property has been in upstream API for at least 13 years so we don't need to worry about it not being supported on older wikis. Thanks.

Issue #76, opened 2015-01-06
Follow "gold standard"


danmichaelo:
Copied the list from https://www.mediawiki.org/wiki/API:Client_code/Gold_standard and started checking off what I think can be checked off.

## Easy to install
- [x] Installation instructions are correct and easy to find
- [x] Library is packaged for installation through appropriate package library (PyPI, CPAN, npm, Maven, rubygems, etc.)
- [ ] Platinum standard: library is packaged for and made available through Linux distributions

## Easy to understand
- [x] Well designed: makes all intended API calls available with the intended level of abstraction with no redundancies
- [ ] Platinum standard: makes the Wikidata API available [#60]

## Well documented
- [ ] Code is commented and readable
- [ ] Documentation is comprehensive, accurate, and easy to find
- [x] Deprecated functions are clearly marked as such
- [ ] Platinum standard: Documentation is understandable by a novice programmer
- [x] Code uses idioms appropriate to the language the library is written in

## Easy to use
- [ ] Has functioning, simple, and well-written code samples for common tasks
- [ ] Demonstrates queries
- [ ] Demonstrates edits

## Handles API complications or idiosyncrasies so the user doesn't have to:
- [x] Login/logout *[No explicit logout method though]*
- [x] Cookies
- [x] Tokens
- [x] Query continuations using the new "continue" and not "query-continue"
- [x] Requests via https, including certificate validation *[at least when closing #70]*
- [ ] Courteous API usage is promoted through code samples and smart defaults
- [x] gzip compression is used by default
- [x] Examples show how to create and use a meaningful user-agent header (as in https://meta.wikimedia.org/wiki/User-agent_policy)
- [ ] Platinum standard: generates a unique user-agent string given name/email address/repository location
- [ ] Efficient usage of API calls
Can be used with the most recent stable version of the language it is written in (e.g. Python 3 compatible)
Easy to debug[rediger | rediger kilde]
Contains unit tests for the longest and most frequently modified functions in the library
Platinum standard: Unit tests for many code paths exist and are maintained
Terrible hacks/instances of extreme cleverness are clearly marked as such in comments
Documentation links to the relevant section/subpage of the API documentation

## Easy to improve
- [x] Library maintainers are responsive and courteous, and foster a thoughtful and inclusive community of developers and users
- [ ] Platinum standard: Project sets clear expectations for conduct[1][2] for spaces where project-related interactions occur (mailing list, IRC, repository, issue tracker). It should:
[...]

PR #279, opened 2022-12-23
Pass status_code to MaximumRetriesExceeded if possible

arbolitoloco1:
if an API request fails and the sleeper exceeds the maximum number of retries, the exception raised doesn't contain any info about the original exception, add the status code when possible

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -409,7 +409,7 @@ class Site(object):
                                 'Retrying in a moment.'
                                 .format(status=stream.status_code,
                                         text=stream.text))
-                    sleeper.sleep()
+                    sleeper.sleep(status_code=stream.status_code)
 
             except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
                 # In the event of a network problem
--- a/mwclient/sleep.py
+++ b/mwclient/sleep.py
@@ -72,6 +72,7 @@ class Sleeper(object):
         with the number of retries.
         Args:
             min_time (int): The minimum sleeping time.
+            status_code (int): Status code of the original exception.
         Raises:
             MaximumRetriesExceeded: If the number of retries exceeds the maximum.
         """
--- a/mwclient/sleep.py
+++ b/mwclient/sleep.py
@@ -66,7 +66,7 @@ class Sleeper(object):
         self.retry_timeout = retry_timeout
         self.callback = callback
 
-    def sleep(self, min_time=0):
+    def sleep(self, min_time=0, status_code=None):
         """
         Sleeps for a minimum of `min_time` seconds. The actual sleeping time will increase
         with the number of retries.
@@ -77,7 +77,7 @@ class Sleeper(object):
         """
         self.retries += 1
         if self.retries > self.max_retries:
-            raise MaximumRetriesExceeded(self, self.args)
+            raise MaximumRetriesExceeded(self, self.args, status_code)
 
         self.callback(self, self.retries, self.args)
 


AdamWill:
I get the idea here, but I don't love the execution. It feels a bit wrong to be passing the status code around like this then just tacking it onto the end of a fairly generic exception.

Couldn't we do this instead?

    diff --git a/mwclient/client.py b/mwclient/client.py
    index c9dfb95..2dea7fd 100644
    --- a/mwclient/client.py
    +++ b/mwclient/client.py
    @@ -13,6 +13,7 @@ from requests_oauthlib import OAuth1
     
     import mwclient.errors as errors
     import mwclient.listing as listing
    +from mwclient.errors import MaximumRetriesExceeded
     from mwclient.sleep import Sleepers
     from mwclient.util import parse_timestamp, read_in_chunks
     
    @@ -409,7 +410,10 @@ class Site(object):
                                     'Retrying in a moment.'
                                     .format(status=stream.status_code,
                                             text=stream.text))
    -                    sleeper.sleep()
    +                    try:
    +                        sleeper.sleep()
    +                    except MaximumRetriesExceeded:
    +                        stream.raise_for_status()
     
                 except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
                     # In the event of a network problem

WDYT?

arbolitoloco1:
Hmmm you're right, that would be def a better solution than mine

Only thing that worries me is that the raised exception would change from `MaximumRetriesExceeded` to whatever requests raises, which would change how exceptions are catched

For example I might want to log when an edit fails so I would do something like
```
try:
    site.edit()
except MaximumRetriesExceeded:
    log_error()
```
which would not work anymore if we raised a different exception

AdamWill:
Erf. The more I look at this the worse it looks, this sleeper design is...kinda funky. I'm looking at how to make it raise something useful on both the 'bad status code' and 'ConnectionError / Timeout' paths, and it feels...icky. I rather want to throw all use of sleeper out of this function and instead just use `urllib3.Retry` instead: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html ...

AdamWill:
But yes, you're right that either of those approaches would effectively be API changes. mmmmph.

AdamWill:
@waldyrious , you have any thoughts here?

waldyrious:
I once [read](https://github.com/JuliaLang/julia/pull/26872#issuecomment-413936868) that the difference between a breaking change and a bugfix is, to some extent, a matter of [interpretation](https://xkcd.com/1172/). This seems like one of those situations, where we may justify the change as a fix to the inconsistent exceptions already raised by the function. But I leave it to your discretion to decide whether people may be relying on the current behavior or not.

More conservatively, what we could do here is cut a new release to include the changes that have already been merged, and then include this improvement (by whichever approach you agree is the most appropriate) as part of a new release with a major version bump.

waldyrious:
Following my comment above, we should probably decide what we want to consider version 1.0 of mwclient to be. IMO that shouldn't refer to a notion of completeness but rather of an API design that we're comfortable as declaring stable and future-proof for a reasonable length of time. I suggest we continue this discussion in a new "Roadmap" issue, or in existing threads like https://github.com/mwclient/mwclient/issues/280#issuecomment-1426485842 or #76.

AdamWill:
Hmm, so after sleeping (hah) on this...one problem with switching to `urllib3.Retry` is it doesn't have callback handling like `Sleeper` does. I don't know if anyone uses that callback feature, but...it's there.

Here's my best effort at a 'clean' improvement here. I wanted to do the handling of MaximumRetriesExceeded just once, and avoid nesting try/except blocks. It doesn't raise anything the function couldn't already raise at least when called with `retry_on_error=False`, but there is a 'behaviour change' to some extent on the `retry_on_error=True` path: you now get an `HTTPError` from `stream.raise_for_status()` in more cases (all non-200 status codes), and you'd get `ConnectionError` or `TimeOut` on that path, not `MaximumRetriesExceeded`. I moved the argument handling at the start out of the try/except block to make it clearer where the handled exceptions can actually happen.

    diff --git a/mwclient/client.py b/mwclient/client.py
    index c9dfb95..a129ddf 100644
    --- a/mwclient/client.py
    +++ b/mwclient/client.py
    @@ -382,22 +382,23 @@ class Site(object):
                                                                 ext=self.ext)
     
             while True:
    -            try:
    -                args = {'files': files, 'headers': headers}
    -                for k, v in self.requests.items():
    -                    args[k] = v
    -                if http_method == 'GET':
    -                    args['params'] = data
    -                else:
    -                    args['data'] = data
    +            toraise = None
    +            wait_time = 0
    +            args = {'files': files, 'headers': headers}
    +            for k, v in self.requests.items():
    +                args[k] = v
    +            if http_method == 'GET':
    +                args['params'] = data
    +            else:
    +                args['data'] = data
     
    +            try:
                     stream = self.connection.request(http_method, url, **args)
    -
                     if stream.headers.get('x-database-lag'):
                         wait_time = int(stream.headers.get('retry-after'))
                         log.warning('Database lag exceeds max lag. '
                                     'Waiting for {} seconds'.format(wait_time))
    -                    sleeper.sleep(wait_time)
    +                    # fall through to the sleep
                     elif stream.status_code == 200:
                         return stream.text
                     elif stream.status_code < 500 or stream.status_code > 599:
    @@ -409,16 +410,29 @@ class Site(object):
                                     'Retrying in a moment.'
                                     .format(status=stream.status_code,
                                             text=stream.text))
    -                    sleeper.sleep()
    +                    toraise = "stream"
    +                    # fall through to the sleep
     
    -            except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
    +            except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as err:
                     # In the event of a network problem
                     # (e.g. DNS failure, refused connection, etc),
                     # Requests will raise a ConnectionError exception.
                     if not retry_on_error:
                         raise
                     log.warning('Connection error. Retrying in a moment.')
    -                sleeper.sleep()
    +                toraise = err
    +                # proceed to the sleep
    +
    +            # all retry paths come here
    +            try:
    +                sleeper.sleep(wait_time)
    +            except errors.MaximumRetriesExceeded:
    +                if toraise == "stream":
    +                    stream.raise_for_status()
    +                elif toraise:
    +                    raise toraise
    +                else:
    +                    raise
     
         def raw_api(self, action, http_method='POST', *args, **kwargs):
             """Send a call to the API."""

AdamWill:
@arbolitoloco1 can you take a look at my alternative PR and see what you think? Thanks! Was kinda waiting for your feedback before merging.

arbolitoloco1:
Sorry for the late response! I just had the time to read it. I think it should work perfectly if as waldyrious said we are fine with doing a breaking change in favor of raising a more specific exception (I'm personally fine with it). `MaximumRetriesExceeded` would still show up in the traceback so that's nice as well.

Thank you very much for your help with the PR! :)

AdamWill:
cool, let's go with that. Thanks!

PR #267, opened 2021-03-02
Improve and fix Site.expandtemplates

KasparNagu:

PR #281, opened 2023-02-09
Fix tests

AdamWill:
Several changes to make the tests work properly.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,4 +23,5 @@ addopts = --cov mwclient test
 [flake8]
 max-line-length = 90
 ignore = 
-	W503  # Line break before binary operator
+    # Line break before binary operator
+    W503
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,0 +1,1 @@
+# This file exists only because pytest-cov generates no coverage information without it.
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -470,7 +470,7 @@ class TestLogin(TestCase):
     def test_clientlogin_success(self, raw_api, site_init):
         login_token = 'abc+\\'
 
-        def side_effect(*args, **kwargs):
+        def api_side_effect(*args, **kwargs):
             if kwargs.get('meta') == 'tokens':
                 return {
                     'query': {'tokens': {'logintoken': login_token}}
@@ -482,9 +482,11 @@ class TestLogin(TestCase):
                     'clientlogin': {'status': 'PASS'}
                 }
 
-        raw_api.side_effect = side_effect
+        raw_api.side_effect = api_side_effect
 
         site = mwclient.Site('test.wikipedia.org')
+        # this would be done by site_init usually, but we're mocking it
+        site.version = (1, 28, 0)
         success = site.clientlogin(username='myusername', password='mypassword')
         url = '%s://%s' % (site.scheme, site.host)
 
@@ -521,6 +523,8 @@ class TestLogin(TestCase):
         raw_api.side_effect = side_effect
 
         site = mwclient.Site('test.wikipedia.org')
+        # this would be done by site_init usually, but we're mocking it
+        site.version = (1, 28, 0)
 
         with pytest.raises(mwclient.errors.LoginError):
             success = site.clientlogin(username='myusername', password='mypassword')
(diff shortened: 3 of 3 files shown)


AdamWill:
Yeah, it looks like Actions aren't turned on for the project yet. If someone could do that it'd be great. I want to merge this so we can get CI for the other pending PRs, it will help with reviewing them.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: Python package with Tox
```
waldyrious, line 2: Let's make this a bit more explicit :)

```suggestion
name: Test the Python package with Tox
```

setup.cfg
```diff
@@ -23,4 +23,4 @@ addopts = --cov mwclient test
 [flake8]
 max-line-length = 90
 ignore = 
-	W503  # Line break before binary operator
```
waldyrious, line 26: According to https://flake8.pycqa.org/en/latest/user/configuration.html, inline comments are not supported, but they can be added in a separate line. Can you do that, please? I'd prefer if we preserved the comment.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py35,py36,py37,py38,flake
+envlist = py27,py35,py36,py37,py38,py39,py310,py311,py312,flake
```
waldyrious, line 2: Would it be worth dropping the [EOL'd Python versions](https://devguide.python.org/versions/)? For the record, I don't have anything against ensuring the project keeps running in old versions, as long as that doesn't compromise our ability to adopt new features or cleanup the codebase.

waldyrious:
I realize the motivation for the `test/__init__.py` file is explained in the commit message (9f0e0adb193786df276087a7139531f5384f5a49), but perhaps it would be convenient to also add a comment to the same effect within the file?

AdamWill:
In general (for your last two notes) I am not caring too hard about cleaning up things like that in this PR because I'm planning to propose much more change later - moving configuration into `pyproject.toml` as far as possible, using diff-cover and diff-quality and black and dropping flake8, using isolated builds in tox...this PR was meant to be more "minimum change necessary to get CI working with the existing stuff so we can review existing PRs and have a base for larger changes".

tox.ini
AdamWill: See comment in main thread - I agree, but kinda planned to do that later. I'm going to suggest dropping Python 2 support entirely, and along with that I'd usually do some stuff like removing any now-unnecessary compat stuff for Python 2 (e.g. use of six), switching to format strings...

AdamWill:
So I don't remember this from the last time I setup Actions, and I can't find a really clear reference, but I *think* when a PR adds a workflow, that workflow won't be run on that PR (which is a bit annoying, other CI systems I've used *do* do this). I think I'll split that commit out into another PR we can merge on its own, and then hopefully we can see how the workflow looks when it runs against this PR.

AdamWill:
https://github.com/mwclient/mwclient/pull/282 has the Actions change.

AdamWill:
OK, looks like the CI is working, whee. (I didn't include Python 2 in the CI stuff since I'm gonna propose dropping it anyway, but I did run it locally and it passes there too). If we can merge this, we can then rebase all other existing PRs and get CI run on them.

tox.ini
AdamWill: gonna mark this as resolved and we'll address it later.

setup.cfg
AdamWill: done.

waldyrious (approved):
LGTM, thanks for taking this on! :tada:

waldyrious:
@AdamWill I merged the PR with the "merge commit" strategy. I think that doing so, while ensuring branches are fully rebased on the tip of `master` prior to the merge, leads to the most readable git history graph, where multiple commits from the same branch are kept atomically separated but still bundled into the same aggregate change unit. In other words, this is the semi-linear git branching strategy. See https://github.com/isaacs/github/issues/1017 for more details and a diagram.

Let me know if you have any thoughts on this matter :)

AdamWill:
Personally I prefer just rebase merging everything for my own projects and that's the most common strategy I run into in other projects, but there's a lot of tastes out there and I don't really mind a whole lot :)

AdamWill:
I tried to merge https://github.com/mwclient/mwclient/pull/267 the way you like, did I get it right?

waldyrious:
Yep, that's perfect, thanks!

PR #283, opened 2023-02-10
Add gh-actions section to tox.ini

AdamWill:
This is a missing bit of the CI enablement, per
https://github.com/ymyzk/tox-gh-actions . I hope this will make the Python 3.6 tests run properly.


AdamWill:
OK, that's good - the tests fail because they need the fixes from https://github.com/mwclient/mwclient/pull/281 , but tox *ran* properly in all envs, which is what I was aiming for with this PR.

PR #285, opened 2023-02-15
Document how to connect using a proxy

### LINKED ISSUE #262: pass a proxy when "getting" and "pushing" files
I am using the mwclient library version 0.10.1 for Python 3.8.3. mwclient.Site() cannot connect to the media wiki site because of the firewall at my company. I have tried setting up a proxy connection to the media wiki site  within the conda Python environment I am using and even in the OS environment variables, but that has not worked. It would be great if a proxy can be passed as an argument to mwclient.Site() when getting and pushing files.

marcfrederick:
Resolves #262 by documenting how to connect to MediaWiki instances using a proxy.

### DIFF
--- a/docs/source/user/connecting.rst
+++ b/docs/source/user/connecting.rst
@@ -55,6 +55,22 @@ Note that MwClient appends its own user agent to the end of your string.
 
 .. _errors:
 
+Using a proxy
+-------------
+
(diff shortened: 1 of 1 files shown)


AdamWill:
LGTM, thanks!

PR #287, opened 2023-02-16
Add a `DeprecationWarning` for `Page.save`

marcfrederick:
`Page.save` is an alias for `Page.edit` meant to provide backwards-compatibility. This change adds a `DeprecationWarning` to the `Page.save` function, to guide users towards the new `Page.edit` function.


AdamWill:
personally I'd be of the mind to just keep it forever and not worry about deprecating/removing it. The only practical reason I can think of to actually remove it would be if we wanted to use the name `save` for something *else*, but that seems unlikely. Maintaining the alias forever would barely count as tech debt, since literally all it has to do is...call the other thing. What does anyone else think?

marcfrederick:
I think it would be a good idea to label the function as deprecated, even if we don't plan on removing it anytime soon. This will make it easier for users to know which of the two functions to use and minimize the impact of a potential removal in the future, should we ever decide to do so.

AdamWill:
for me it just feels like unnecessarily annoying people with a DeprecationWarning if there's no real reason to ever actually remove it. especially for users of some client of a client of mwclient who can't necessarily change it themselves, who are just stuck looking at ugly warnings until a maintainer makes a change that isn't really necessary.

marcfrederick:
I was under the assumption the `DeprecationWarning` would be hidden, but just saw that we're explicitly making it visible in the `__init__.py`.

https://github.com/mwclient/mwclient/blob/209568f38199973ef536d17adc64bd262795bf26/mwclient/__init__.py#L31-L32

With that in mind, I have to agree with you on just keeping the alias and will close this PR. Thanks for the feedback :)

PR #288, opened 2023-02-16
Add/Update some documentation

marcfrederick:
Updates the documentation of some functions in the `Site` class. Additionally:

* Updates the `basic_edit.py` example to use `page.edit` instead of the older `page.save`
* Adds a `retry_on_error=True` argument to `Site.raw_api`. This argument already existed, but was hidden behind the `**kwargs`. Making it explicit will help users discover that option.


mwclient/client.py
```diff
@@ -344,7 +422,7 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
         """
         Perform a generic request and return the raw text.
 
-        In the event of a network problem, or a HTTP response with status code 5XX,
+        In the event of a network problem, or an HTTP response with status code 5XX,
```
AdamWill, line 425: ❤️

mwclient/client.py
```diff
@@ -421,7 +499,19 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
...
+            action (str): The MediaWiki API action to perform.
+            http_method (str): The HTTP method to use in the request.
+
+        Returns:
+            The API response.
+
+        Raises:
+            errors.APIDisabledError: The MediaWiki API is disabled for this instance.
+            errors.InvalidResponse: The API response could not be decoded from JSON.
+        """
```
AdamWill, line 514: it can raise various other things, via `raw_call`: at least `requests.exceptions.HTTPError` and `mwclient.errors.MaximumRetriesExceeded`.

mwclient/client.py
```diff
@@ -421,7 +499,19 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
                 sleeper.sleep()
 
     def raw_api(self, action, http_method='POST', *args, **kwargs):
-        """Send a call to the API."""
+        """Send a call to the API.
+
+        Args:
+            action (str): The MediaWiki API action to perform.
+            http_method (str): The HTTP method to use in the request.
```
AdamWill, line 506: as well as `action` and `http_method`, this effectively takes `retry_on_error` and any args that `_query_string` accepts (since it passes `*args` and `**kwargs` to `_query_string`).

mwclient/client.py
```diff
@@ -440,13 +530,38 @@ def raw_api(self, action, http_method='POST', *args, **kwargs):
...
-        """Sends a call to index.php rather than the API."""
+        """Sends a call to index.php rather than the API.
+
+        Args:
+            action (str): The MediaWiki API action to perform.
+            http_method (str): The HTTP method to use in the request.
+
+        Returns:
+            The API response.
+        """
```
AdamWill, line 541: as with `raw_api`, this can raise anything `raw_call` can. though interestingly there's no way to set `retry_on_error` false here - you're stuck with it set true.

mwclient/client.py
```diff
@@ -440,13 +530,38 @@ def raw_api(self, action, http_method='POST', *args, **kwargs):
...
+            revision (int): The required revision.
+            raise_error (bool): Whether to throw an error if the version of the current
+                wiki is below the required version. Defaults to `True`.
+
+        Returns:
+            `False` if the version of the current wiki is below the required version, else
+                `True`. If `raise_error=True` then nothing is returned.
+
+        Raises:
+            errors.MediaWikiVersionError: The thw current wiki is below the required
```
AdamWill, line 562: "thw" here seems like a typoed duplicate of "The". Should be omitted.

This can also raise `NotImplementedError` if the `revision` arg is set. Which is...weird. Why does it do that? Bit of a footgun.

mwclient/client.py
```diff
@@ -440,13 +530,38 @@ def raw_api(self, action, http_method='POST', *args, **kwargs):
...
+        Args:
+            major (int): The required major version.
+            minor (int): The required minor version.
+            revision (int): The required revision.
+            raise_error (bool): Whether to throw an error if the version of the current
+                wiki is below the required version. Defaults to `True`.
+
+        Returns:
+            `False` if the version of the current wiki is below the required version, else
+                `True`. If `raise_error=True` then nothing is returned.
```
AdamWill, line 559: It also returns nothing if the site is not yet initialized and `raise_error` is `None` ( but not if `raise_error` is anything *other* than `None`, including `False`, which is odd). This whole method is *bizarre*. PR time!

mwclient/client.py
AdamWill: It also raises `RuntimeError` if the site is not initialized and `raise_error` is anything but `None`.

AdamWill:
aside from those notes, this looks great! Thanks a lot for working on it. sorry for all the individual comments, I thought I only had one or two notes but kept seeing more things :)

mwclient/client.py
marcfrederick: I've pushed an updated docstring, adding `errors.MaximumRetriesExceeded`, `requests.exceptions.HTTPError`, `requests.exceptions.ConnectionError`, and `requests.exceptions.Timeout`.

mwclient/client.py
marcfrederick: I've added the `NotImplementedError` and `RuntimeError` to the docstring, this method is definitely a bit weird :)

```
RuntimeError: It `raise_error` is `None` and the `version` attribute is unset
    This is usually done automatically on construction of the :class:`Site`,
    unless `do_init=False` is passed to the constructor. After instantiation,
    the :meth:`~Site.site_init` functon can be used to retrieve and set the
    `version`.
NotImplementedError: If the `revision` argument was passed. The logic for this
    is currently unimplemented.
```

mwclient/client.py
marcfrederick: I've updated the `Returns:` section:

```
`False` if the version of the current wiki is below the required version, else
`True`. If either `raise_error=True` or the site is uninitialized and
`raise_error=None` then nothing is returned.
```

mwclient/client.py
marcfrederick: Added the exceptions from `raw_call` here.

mwclient/client.py
marcfrederick: What do you think about just adding an explicit `retry_on_error=True` to the function signature here? This should be a non-breaking change, as it merely makes a kwarg explicit, and makes the intention obvious to users.

AdamWill: yeah, that seems pretty reasonable to me.

marcfrederick: Alright, I've added a new commit to this PR making that change. Thanks :)

AdamWill:
Awesome, thanks again for working on this.

Issue #290, opened 2023-02-19
KeyError: 'expiry' missing in `page.info.protection`


yzqzss:
If a protected page is without the 'expiry' key in its meta-info, mwclient will crash.

See also: https://github.com/mediawiki-client-tools/mediawiki-scraper/issues/131

---

info["protection"]: (normal protected page)
```
[OrderedDict([('type', 'edit'), ('level', 'sysop'), ('expiry', 'infinity')]), OrderedDict([('type', 'move'), ('level', 'sysop'), ('expiry', 'infinity')])]
```

info["protection"]: ([error page sample])
```
[OrderedDict([('type', 'edit'), ('level', 'sysop')]), OrderedDict([('type', 'move'), ('level', 'sysop')])]
```

>  Protected "System Recovery Status" ([Edit=Allow only administrators] (_indefinite_) [Move=Allow only administrators] (_indefinite_))

[error page sample]: https://www.linux-mips.org/wiki?title=System_Recovery_Status&oldid=66035

## How to reproduce:

```
session = requests.Session()
session.verify = False
requests.packages.urllib3.disable_warnings()
site = mwclient.Site("www.linux-mips.org", "/mediawiki/", "https", pool=session)
for page in site.allpages(namespace=0):
    pass
```

marcfrederick:
There's definitely something weird going on here. There's another protected page on that wiki, for which the `expiry` is present.

```
$ curl -sk 'https://www.linux-mips.org/mediawiki/api.php?generator=allpages&gapnamespace=0&gapfilterredir=all&gapdir=ascending&gapfilterlanglinks=all&gaplimit=500&prop=info%7Cimageinfo&inprop=protection&iiprop=timestamp%7Cuser%7Ccomment%7Curl%7Csize%7Csha1%7Cmetadata%7Carchivename&continue=&meta=userinfo&uiprop=blockinfo%7Chasmsg&action=query&format=json' | jq -c '.query.pages[].protection | select(length > 0)'
[{"type":"edit","level":"sysop","expiry":"infinity"},{"type":"move","level":"sysop","expiry":"infinity"}]
[{"type":"edit","level":"sysop"},{"type":"move","level":"sysop"}]
```

In my opinion, we should still try to handle this case gracefully (e.g. by falling back to `None` as done in #291) instead of throwing an exception. The current behavior makes instantiating such pages straight up impossible, which is very unexpected from the perspective of an end-user (who typically has no influence on the configuration of the wiki).

[...]

PR #291, opened 2023-02-19
Handle page protection having no expiry (#290)

AdamWill:
Issue #290 gives an example of a wiki with protected pages whose protection definitions have no 'expiry' key at all. This seems strange and, on a quick through the mediawiki code, difficult to achieve, but since there's a live site out there that does it, and mediawiki *does* seem to have at least some code to handle such cases (e.g. how the expiry shows as 'indefinite' in the page information), let's handle it too, by representing this as None.

### DIFF
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -53,7 +53,7 @@ class Page(object):
         self.exists = 'missing' not in info
         self.length = info.get('length')
         self.protection = {
-            i['type']: (i['level'], i['expiry'])
+            i['type']: (i['level'], i.get('expiry'))
             for i in info.get('protection', ())
             if i
         }
--- a/test/test_page.py
+++ b/test/test_page.py
@@ -165,6 +165,12 @@ class TestPage(unittest.TestCase):
         assert page.can('edit') is True  # User has 'autoconfirmed'  right
         assert page.can('move') is True  # User has 'sysop' right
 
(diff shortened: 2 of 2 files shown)


AdamWill:
We kinda have, well, four choices here, I guess:

1. Represent no expiry as `None` (my initial approach here)
2. Represent no expiry as `""`
3. Represent no expiry as `"infinity"`
4. Raise an exception for no expiry (accidentally as we are at present, or explicitly)

1 and 2 give the mwclient user the information that expiry is not explicitly set, it's just a question of whether we want it to always be a string type, or use the more explicit `None`. 3 hides that information from the user but perhaps makes things a bit more convenient (one less potential value for a user to care about). 4 is the "no, this really SHOULD be set, your wiki is broken" option, I guess.

What does anyone else think about this? I'm not really sure which of 1, 2 or 3 would be best. It probably is better to handle this *somehow* rather than go with option 4, though, I guess.

marcfrederick:
I would favor your first approach (fallback to `None`) here, as it is the clearest representation of what was returned by the API. Thinking about types (should we ever decide to add them), using an `Optional[str]` instead of a `str` here, makes this error case obvious to users, who can then make a decision whether to handle it (or not).

marcfrederick (approved):

AdamWill:
OK, let's go with that.

PR #303, opened 2023-07-31
Do not append custom name to API

Krummers:
This prevents usage from Wikis that check users' APIs in a way. CloudFlare is an example.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -140,7 +140,7 @@ class Site(object):
 
             # Set User-Agent header field
             if clients_useragent:
-                ua = clients_useragent + ' ' + USER_AGENT
+                ua = clients_useragent
             else:
                 ua = USER_AGENT
             self.connection.headers['User-Agent'] = ua


marcfrederick:
The `mwclient/0.10.0 (https://github.com/mwclient/mwclient)` part that's appended to the User-Agent is required by the [Wikimedia User-Agent policy](https://meta.wikimedia.org/wiki/User-Agent_policy):

```
<client name>/<version> (<contact information>) <library/framework name>/<version>
```

If this is causing problems we should, in my opinion, add a parameter (something like `clients_useragent_library_suffix_enabled`) to the `Site` object that toggles this suffix (default `True`) instead of outright removing it.

Krummers:
I will see if I can add this parameter in the next week or so.

marcfrederick:
@Krummers I took another look at how we currently implement this and I think you should be able to set a fully custom User-Agent without making changes to the library by updating the headers of the underlying session.

```python
site = mwclient.Site()
site.connection.headers["User-Agent"] = "SomeUserAgent/1.0"
```

This should override the default User-Agent set in the constructor of the `Site` class.

https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/client.py#L142-L146C13

Krummers:
If this is possible, would this new feature still be necessary?

marcfrederick:
If this works for you then in my opinion not. We can get back to this in the future if we get more similar reports.

Krummers:
After finally picking up my project again after a couple of weeks, I discovered the method you provided does not actually work, since the correct user agent would need to be provided at the same time the object is created.

This does work:
```
import mwclient as mc
wiki = mc.Site("custom.wiki.com", custom_headers = {"User-Agent": "SpecificUserAgent"})
```

Issue #302, opened 2023-07-14
Force users to respect Wikipedia's User Agent policy


XioNoX:
(I'm a SRE at the Wikimedia Foundation)

We regularly get contacted by people getting rate limited while using the API, either directly or for example using your library.
Our API also gets occasionally and accidentally abused.

To help with those 2 issues, we require API users to define a non-default User-Agent, explained in this document: https://meta.wikimedia.org/wiki/User-Agent_policy

I'm wondering if it would be possible to add documentation in this regard to your library, as well as make the UA a mandatory configuration option.

You can see an example implementation on https://github.com/martin-majlis/Wikipedia-API/issues/63

Thanks

marcfrederick:
I've added a new warning to the documentation and updated the initial examples/quickstart to set the `clients_useragent` in #306. I'm personally undecided about whether the user agent should be a mandatory field, as this would introduce a breaking change to the library and is not something that's actually required on all wikis.

AdamWill:
Yeah, I have the same hesitation. For instance, my own use of mwclient is for the Fedora wiki, not Wikipedia, and that does not require a user agent.

XioNoX:
Thanks for your answers!
It would be interesting to know the feedback from Wikipedia-API about the breaking change. But as it's something easy to implement on the user side, I don't see it as a blocker (doesn't require to do significant code change to be compliant).
From experience, users often don't follow recommendations from the documentation (at best do copy/paste).
Even if a wiki doesn't require a specific user agent, it seems to me that setting one is being "a good internet citizen".
To ease the breaking change side, one option could be to show a warning (like Python's famous `DeprecationWarning`) for some time.
Whatever direction you're going, thank you for looking and thinking about this ticket, it's appreciated.

martin-majlis:
I have made the change in June 2021. ChatGPT has knowledge cut-off in September 2021 and is not aware of this backward incompatible change. 🙈 I should have made that change immediately. :) 

I have at least made the error messages more user friendly - https://github.com/martin-majlis/Wikipedia-API/pull/256.

andrei-korshikov:
https://meta.wikimedia.org/wiki/User-Agent_policy reads:

> This page has been moved to "[Policy:User-Agent policy](https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_User-Agent_Policy)"

I think that link in
[...]

PR #306, opened 2023-09-05
Update docs to recomment setting a user-agent

marcfrederick:
Improve the documentation in response to #302 by adding a new warning to the documentation more clearly recommending users to set a User-Agent.

This also updates the initial examples on the "Connecting to your site", "Page operations", as well as "Quickstart" pages with a User-Agent.


AdamWill:
LGTM, thanks!

PR #307, opened 2023-09-05
Add .readthedocs.yaml configuration file

marcfrederick:
Read the Docs has started requiring a .readthedocs.yaml configuration file to be present in the repository in order to build the documentation.

https://blog.readthedocs.com/migrate-configuration-v2/

According to the timeline we have until 2023-09-25 to migrate to the new configuration file, but we've already had some builds fail during their 48-hour brownout on 2023-09-04.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,15 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
+  configuration: docs/source/conf.py
+
+python:
+  install:
+    - method: pip
+      path: .
+    - requirements: docs/requirements.txt
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,2 @@
+sphinx==7.2.5
+sphinx_rtd_theme==1.3.0


AdamWill:
I tend to hate hardcoded versions which inevitably go hilariously stale over the years...but at a quick look it seems like there aren't any 'latest' or 'stable' redirects we can use for these, so we're just stuck. :(

marcfrederick:
Yep, we're pretty much stuck with this -- the setup is more or less copy-paste. Only change I made was update the versions in the `docs/requirements.txt` to the current versions.

waldyrious:
FYI, I recently got an email about our ReadTheDocs integration — it seems to be for a separate issue than what's handled in this PR (I've forwarded it to you both so you can check), but I thought I'd take the opportunity to mention publicly that besides access to the `mwclient` GitHub organization and the `mwclient` PyPI project, it's also necessary to ensure maintainers have access to the `mwclient` project on ReadTheDocs.

Hopefully we will later create a proper documentation for project maintenance, mentioning these account access steps that need to be performed when adding new maintainers, the [preferred PR merge strategy](https://github.com/mwclient/mwclient/pull/310#issuecomment-1913399249), etc.; but in the meantime, hopefully this comment will offer something searchable.

PR #308, opened 2023-09-06
Fix inheritance issue in __repr__() methods

### LINKED ISSUE #309: Hardcoded class names in __repr__() methods
Hardcoded class names are a violation of the [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle). Use `.__class__.__name__` instead.

That said, the following lines need to be modified:

https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/client.py#L271-L272
https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/image.py#L77-L78
https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/listing.py#L128-L129
https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/listing.py#L204-L205
https://github.com/mwclient/mwclient/blob/afeb2a2f8bdb4fb038dff250802ab0b58a627d68/mwclient/page.py#L88-L89

NguoiDungKhongDinhDanh:
Fixes #309.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -269,7 +269,7 @@ class Site(object):
     }
 
     def __repr__(self):
-        return "<Site object '%s%s'>" % (self.host, self.path)
+        return "<%s object '%s%s'>" % (self.__class__.__name__, self.host, self.path)
 
     def get(self, action, *args, **kwargs):
         """Perform a generic API call using GET.
--- a/mwclient/image.py
+++ b/mwclient/image.py
@@ -75,4 +75,8 @@ class Image(mwclient.page.Page):
             return self.site.connection.get(url).content
 
     def __repr__(self):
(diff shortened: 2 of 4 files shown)


marcfrederick:
Looks good to me. I think this *technically* [breaks backwards compatibility](https://xkcd.com/1172/) as the `__repr__` of sub-classes changes, though that should be pretty much irrelevant.

If none of the other maintainers are opposed then I'll merge this in a few days.

AdamWill:
LGTM too. I'm willing to take the bet that this won't break anyone in the real world...we should definitely mention it in the notes for the next release, though.

PR #311, opened 2023-09-28
Convert setup.py to pyproject.toml

NguoiDungKhongDinhDanh:
pyproject.toml is pretty much the standard nowadays I believe. This, among others, allows us to just run `pytest` instead of `python setup.py pytest` which is a handful.

I'm not sure if this breaks anything though.


NguoiDungKhongDinhDanh:
I don't know why 3.6 is failing. [The URL](https://tox.readthedocs.io/en/latest/example/package.html) in [the log message](https://github.com/mwclient/mwclient/actions/runs/6333929068/job/17202769180?pr=311) leads to nowhere.

AdamWill:
this is kind of overly aggressive, I think. for my own projects for now I prefer to take a dual approach, providing both `pyproject.toml` and `setup.py` such that either can be used alone.

AdamWill:
also it would probably make most sense to do this after https://github.com/mwclient/mwclient/pull/310 and drop the python 2.x bits from it - this doesn't have a snowball's chance in hell of working with python 2.x, after all.

AdamWill:
If you compare the 3.6 and 3.7 logs carefully, the 3.6 env just isn't really parsing the deps from pyproject.toml at all, it never tries to pull in requests-oauthlib . That's why the tests fail.

3.6 and 3.7 are EOL now *officially*, of course. But RH, SUSE and Debian are all still supporting releases with system Python 3.6 or earlier. Debian 8 (Jessie) has 3.4. RHEL (disclaimer: I work for RH) *7* is still in extended support, and its system Python is 2 (ugh). Even if we disregard that, RHEL 8's system Python is 3.6. Best as I can tell for SUSE, SLES 12 is still in support and has system Python 3.4, with a 3.6 package available. (For Ubuntu, 20.04 is the oldest release still in LTS, and it seems to have 3.8).

We don't *have* to care about ancient LTS distros on our master branch, of course, but...it's considerations like that which make me reluctant to go pyproject.toml-only on an existing project (I probably would on a new one).

NguoiDungKhongDinhDanh:
Just to be clear: This is not meant to be merged on the fly; I just feel that this needs some place in the PR tab. Until we eventually get to the point where we can ignore anything lower than 3.6, that is.

AdamWill:
We're now officially baselined at 3.6, so we could maybe take another look at this, if you can get it to work on 3.6.

marcfrederick:
This has been superseded by #360, which included the commits from this PR. Thank you for the work you put into this, @NguoiDungKhongDinhDanh.

Issue #315, opened 2024-01-27
httpauth only allows latin-1 characters in username and password


LaoshuBaby:
Hello, today I am trying to use mwclient to edit Wikipedia. I want to test whether this framework is functional.

My code are show below:

```
user_agent = "laoshubaby's robot"
site = mwclient.Site(
    "zh.wikipedia.org",
    # path="/",
    clients_useragent=user_agent,
    httpauth=("快乐的老鼠宝宝", "my_password_in_plain_text")
)
# site.login("快乐的老鼠宝宝", "my_password_in_plain_text")


page = site.pages["User:快乐的老鼠宝宝/沙盒"]
text = page.text()
add_text = "是鼠宝宝的姬器人！"
page.edit(text + add_text, add_text)
```

If I try to login with `httpauth`` parameter, this will cause a error like:

```
PS C:\Programmer\learn_python> python wiki.py
Traceback (most recent call last):
  File "C:\Programmer\learn_python\wiki.py", line 37, in <module>
    site = mwclient.Site(
           ^^^^^^^^^^^^^^
  File "C:\Environment\Python\Python312\Lib\site-packages\mwclient\client.py", line 130, in __init__
    self.site_init()
  File "C:\Environment\Python\Python312\Lib\site-packages\mwclient\client.py", line 150, in site_init
    meta = self.get('query', meta='siteinfo|userinfo',
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Environment\Python\Python312\Lib\site-packages\mwclient\client.py", line 234, in get
    return self.api(action, 'GET', *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Environment\Python\Python312\Lib\site-packages\mwclient\client.py", line 285, in api
    info = self.raw_api(action, http_method, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Environment\Python\Python312\Lib\site-packages\mwclient\client.py", line 437, in raw_api
    res = self.raw_call('api', data, retry_on_error=retry_on_error,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #316, opened 2024-01-27
HTTP basic auth: encode username and password as UTF-8 (#315)

AdamWill:
As discussed upstream in
https://github.com/psf/requests/issues/4564 , HTTP basic auth usernames and passwords sent to requests as Python text strings are encoded as latin1. This of course makes it impossible to log in with a username or password containing characters not represented in latin1, as the reporter of #315 found out.

To work around this rather old-fashioned default, let's intercept string usernames and passwords and encode them as utf-8 before sending them to requests.

Anyone dealing with a really old server that can't handle utf-8, or something like that, can encode the username and password appropriately and provide them as bytestrings.


AdamWill:
Update - rebased on the "drop Python 2.x support" PR I just merged.

marcfrederick (approved):

PR #317, opened 2024-01-27
Update GHA config for Python 3.12 final

AdamWill:
I think we can just do this now Python 3.12 is stable.


AdamWill:
hmm. Don't know why we didn't get a 3.12 run on the PR, but I *think* it's because of the mismatch between the PR and the base branch, this really ought to work AFAICS. We can try merging it and see if another PR behaves as expected?

marcfrederick (approved):

Issue #280, opened 2023-02-07
Does this project need a fork or new maintainers?


WolfgangFahl:
There has been no new release since 2020-05. I am currently having trouble with using it with Mediawiki 1.39 and before i am adding another issue I'd love to know whether this is worthwhile and will lead to any resolution.

As of 2023-02-07 there are 9 open Pull requests and 36 open issues. I don't see an milestones configure and there is no open discussion. 

**Who could change this situation and how?**

waldyrious:
Hi Wolfgang — thanks for raising this issue, and for your interest in the project :)

For the record, I do have administrative permissions in the project, and am aware of the issues and pull requests that have been submitted over time, but I don't have sufficient knowledge of the codebase to be able to make technical decisions regarding them. For what it's worth, whenever a decision has been within the bounds of my ability or confidence level, I did merge the relevant patches (see e.g. https://github.com/mwclient/mwclient/pull/242#issuecomment-924940825, https://github.com/mwclient/mwclient/pull/261, and https://github.com/mwclient/mwclient/pull/269).

That said, I definitely agree we need additional maintainers, but unfortunately I'm not aware of any candidates. I'd be happy to grant collaborator rights to people who make a compelling case for having both the willingness to maintain the project and the technical know-how to do so well (e.g. one of the [past contributors](https://github.com/mwclient/mwclient/graphs/contributors) to the project). Speaking of which, besides @danmichaelo and @btongminh, who are currently inactive, the people with the most commits to the repo are @lukasjuhrich, @marcfrederick and @AdamWill. Pinging them in case they might be interested.

By the way, I will attend this year's Wikimedia Hackathon, so I will also [make it one of my goals there](https://www.mediawiki.org/w/index.php?title=Wikimedia_Hackathon_2023/Connect&diff=prev&oldid=5763689) to scout the Wikimedia technical community for possible maintainers to this project (it would be especially helpful to get the ability to publish releases in PyPI).

Let me know if there's anything else you can think of that would help!

waldyrious:
Hey! Thanks for chiming in :) and quite quickly, too!

It seems to me you know the internals of the project better than I do, at least, so having you as part of the maintenance team would already be an improvement! :smile: Besides, your past contributions have been generally well received and received little to no requested changes, which is also a very good sign that your understanding of the project is aligned with what its vision has been so far.

I will add you as a member of the organization. At the very least, that will increase our bus number by 1, so it seems justified.

[...]

PR #319, opened 2024-01-27
Use Coveralls Github Action to send data to coveralls again

AdamWill:
This was lost when Travis stopped working, so our coveralls data is frozen in 2021. This should get it working again using https://github.com/marketplace/actions/coveralls-github-action if I got it right.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,7 +1,7 @@
 ---
 name: Test Python package with Tox
 
-on: [pull_request]
+on: ["push", "pull_request"]
 
 jobs:
   build:
@@ -34,6 +34,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox tox-gh-actions
+          pip install tox tox-gh-actions coverage
       - name: Test with tox
         run: tox
+      - name: Produce coverage.lcov for coveralls
+        if: matrix.python-version == '3.11'
+        run: coverage lcov
+      - name: Coveralls
+        if: matrix.python-version == '3.11'
+        uses: coverallsapp/github-action@v2


AdamWill:
hum, we don't really want to do it seven times per PR. I'll have to tweak this a bit.

AdamWill:
whew, well that syntax is surprisingly finicky - the quotes *have to* be single, YAML thing I guess? - but it works now.

AdamWill:
The change to run the workflow on push as well as pull request I think is necessary for the badge coverage number to be updated (it seems to only be updated when coveralls calculates coverage of the master branch, not when it calculates coverage for a PR).

I'm trying to get the thing where coveralls posts a comment on the PR to work, but it doesn't seem to be yet. I need to check a setting on coveralls.io but I don't have access to the project settings there, something to do with permissions...trying to figure it out...

waldyrious:
> I'm trying to get the thing where coveralls posts a comment on the PR to work, but it doesn't seem to be yet. I need to check a setting on coveralls.io but I don't have access to the project settings there, something to do with permissions...trying to figure it out...

@AdamWill I got a _"Third-party application approval request for `mwclient`"_ email from GitHub asking to grant Coveralls access to the repository, and have accepted it. Hopefully that unblocks the issue :)

In any case, [I've made you an owner](https://github.com/mwclient/mwclient/issues/280#issuecomment-1913388227) in the mwclient GitHub organization, so I suppose you should now have the same access levels as I do. In particular, you should be able to access the application access configurations at https://github.com/organizations/mwclient/settings/oauth_application_policy, and should probably be able to access <https://coveralls.io/github/mwclient/mwclient/settings> as well. Please confirm!

AdamWill:
Indeed I can, but I've set all the things that I can find right, and coveralls *still* isn't commenting. I've sent a mail to coveralls support, maybe they can help. :D

AdamWill:
So Coveralls support got back with a very detailed reply:

"Apologies for the delay.

It sounds like you're all set up to receive PR Comments, so, as I suspected, I see the issue has a different cause, which is that Coveralls can't find the base build for your PR build, which it needs to derive a coverage change calculation, which is the content of any PR Comment it would send. (We don't do a great job of letting you know we couldn't send one, right now. We are working on that.)

The clue that this is the case, is that your PR Build reads "FIRST BUILD" in the Build Info table at the top of its Build Page. This sometimes happens when users are getting set up: They've added all the Coveralls configuration details to their PR, but they haven't merged those changes into their default branch yet. Since the base build of a PR build will almost always be a `push` build on your default branch, there is no base build for that new PR yet.

That's usually fixed by merging the PR changes into your default branch, at which point, we start getting coverage reports for your `push` builds on master, and, so, your next PR build(s) have base builds and can receive PR Comments.

I'll leave it there since I'm almost certain that the changes related to switching from Travis-to-GitHub-Actions have created this situation for you, but if for some reason, you take those next steps and still don't receive PR Comments, please let me know and we'll take it from there."

so, tl;dr, it sounds like if we go ahead and merge this, we *should* then get comments as expected on all other PRs. Are other maintainers OK with that? If nobody yells, I'll go ahead and merge this in a few days and we'll see what happens.

AdamWill:
If I'm reading that right, possibly we'll need to merge this and then do one *more* push to master somehow before comments start working. Not totally sure if the merge of this PR would give us the initial push build on master that coveralls needs.

marcfrederick (approved):

PR #321, opened 2024-01-28
Remove some Python 2.x leftovers

marcfrederick:
* `Iterator.next()` has been replaced with `Iterator.__next__()`
* The `object.__unicode__()` method is not used anymore
* Default source encoding has changed from ASCII to UTF-8


AdamWill:
For some reason I don't have a "review changes" button, but LGTM, thanks!

PR #324, opened 2024-02-09
Update authentication method docs

### LINKED ISSUE #322: Update authentication documentation page
The info on authenticating in the docs doesn't cover `clientlogin()`, and - as suggested in https://github.com/mwclient/mwclient/issues/315 - could make it clearer that the options that allow HTTP authentication aren't alternatives for logging into mediawiki itself.

marcfrederick:
Attempt to resolve #322, by clarifying the authentication section of the documentation.

### DIFF
--- a/docs/source/user/connecting.rst
+++ b/docs/source/user/connecting.rst
@@ -115,8 +115,8 @@ unauthenticated, just set
 
 .. _oauth:
 
-OAuth
-^^^^^
+OAuth Authentication
+^^^^^^^^^^^^^^^^^^^^
 
 On Wikimedia wikis, the recommended authentication method is to authenticate as
 a `owner-only consumer`_. Once you have obtained the *consumer token* (also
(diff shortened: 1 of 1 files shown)


AdamWill:
I kinda preferred how the existing file has OAuth *first*. We really want to encourage people to use OAuth for what's probably the most common use case (connecting to wikimedia wikis). In my experience you should always put the thing people are usually meant to do *first*, then other stuff comes afterwards.

marcfrederick:
> In my experience you should always put the thing people are usually meant to do _first_, then other stuff comes afterwards.

Good point, I've switched the sections back to the original order.

AdamWill:
Should we also document `clientlogin()`?

AdamWill:
Well, this is clearly an improvement anyway, so let's go with it. I guess I'll write a follow-up to document `clientlogin`.

PR #325, opened 2024-02-10
Update PyPi classifiers

marcfrederick:
Add missing classifiers for Python versions >= 3.9


AdamWill:
this looks nice and trivial as our 'extra commit' that coveralls might need, so let's merge it!

Issue #284, opened 2023-02-11
mwclient 1.0 release


RheingoldRiver:
Following up from [this commnet](https://github.com/mwclient/mwclient/pull/279#issuecomment-1426669958), what should a 1.0 release of mwclient look like?

Things to consider:

- The "gold standard" for MediaWiki libraries, see #76 
    - Biggest thing missing seems to be documentation
- long-term stability of usage, even if implementation will change

One other thing I'd like to consider is that Discord bots are becoming more and more popular, with discord.py and discord.js being the two main libraries. mwclient should be the go-to library for dpy, however, we don't have any async capability making it ill-suited for discord. Maybe we can add async functions for at least login, read, edit? [This client](https://github.com/Wadu436/aiomwclient) does exist and I think it would be great for it to be merged into mwclient proper. I just spoke to @Wadu436 and he's interested in at least discussing the issue, even if not necessarily contributing code. It's possible that it doesn't make sense to combine these into one library; however, if they remain separate projects I think one is more likely to go unmaintained, which I think is unfortunate.

AdamWill:
I'd say we de facto have "long-term stability of usage" - anyone using mwclient now has to be pretty used to how it works at present =) I guess I wouldn't want to make any overly radical changes. Improving documentation would definitely be a good idea.

I have never written much async stuff and don't have any interest in discord bots myself, so I'm not sure I have a lot of opinions about that side. It seems like aiomwclient is basically....mwclient, only with `async def` everything? As I said, I'm not that familiar with async, so...how disruptive would it be to merge that kind of thing into mwclient proper, but preserving the synchronous interface for users who have no interest in async usage?

RheingoldRiver:
One thing to consider regarding stability is that in a 1.0 release, if we retain `page.save`, we will certainly need to keep it forever - currently both `page.save` and `page.edit` do the same thing. IMO there is no question that we should keep both (I've personally always used `page.save`) but if there's some reason to use `edit` only, this change should be made in 1.0 I think.

I'm not sure if there's any other similar aliases that exist.

See: https://github.com/mwclient/mwclient/blob/b1eb9772f6c7e3a981b789168e2596ca7820495a/mwclient/page.py#L177

Wadu436:
> It seems like aiomwclient is basically....mwclient, only with async def everything? 

[...]

PR #327, opened 2024-02-17
Implement error handling for MediaWiki warnings on existing files

### LINKED ISSUE #211: site.upload() should raise exception for warnings
If you call `wiki.upload()` with an existing filename, MediaWiki returns a warning and does not update the file. Since mwclient swallows these warnings, I just spent a lot of time debugging my program before I figured out what was going on. These warnings should probably be raised as errors instead of ignored.

marcfrederick:
Fixes #211 by raising an error when `ignore=False` and the MediaWiki response contains the `upload.warnings.exists` key. Previously, this warning was logged, but the call to `Site.upload` succeeded. This could be easily missed by users, especially when they haven't configured the log level.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -904,6 +904,7 @@ class Site:
         Raises:
             errors.InsufficientPermission
             requests.exceptions.HTTPError
+            errors.FileExists: The file already exists and `ignore` is `False`.
         """
 
         if file_size is not None:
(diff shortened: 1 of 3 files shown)


AdamWill:
LGTM, @RheingoldRiver also acked it in #284 so let's go ahead and merge. Thanks!

PR #331, opened 2024-05-04
Add an automated release workflow using Github Actions

AdamWill:
This adds a release workflow, and updates tool config a bit to make the latest incarnation of bumpversion happy. I've tested that you can do `bump-my-version bump major` on top of this and it looks like it does everything we'd want - update all the appropriate version strings and create a tagged commit. So I *hope* that, once we merge this, we can submit another PR created by editing `CHANGELOG.md` then running `bump-my-version bump major`, then merge that PR and the release should happen magically.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: Build and publish Python distribution to PyPI and TestPyPI
+
+on: push
+
+jobs:
+  build:
+    name: Build distribution
+    if: startsWith(github.ref, 'refs/tags/v') # only publish to PyPI on version tag pushes
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v5.1.0
+      with:
+        python-version: "3.x"
+
+    - name: Install pypa/build
+      run: python3 -m pip install build --user
+    - name: Build a binary wheel and a source tarball
+      run: python3 -m build
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v4
+      with:
+        name: python-package-distributions
+        path: dist/
+
+  publish-to-pypi:
+    name: Publish Python distribution to PyPI
+    if: startsWith(github.ref, 'refs/tags/v') # only publish to PyPI on version tag pushes
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/mwclient
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
+
+    steps:
+    - name: Download all the dists
+      uses: actions/download-artifact@v4
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Publish distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -109,3 +109,18 @@ When it is ready, push your branch to your remote:
 
 Then you can open a pull request on GitHub. You should see a URL to do this
 when you push your branch.
+
+Making a release
+----------------
+
+These instructions are for maintainers of the project.
+To cut a release, ensure ``CHANGELOG.md`` is updated, then use
+`bump-my-version <https://callowayproject.github.io/bump-my-version/>`_:
+
+.. code:: bash
+
+    $ pip install bump-my-version
+    $ bump-my-version bump major|minor|patch
+
+Then check the commit looks correct and is tagged vX.Y.Z, and push. The
+``.github/workflows/release.yml`` action will publish to PyPI.
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -12,12 +12,14 @@ Development environment
 
 If you plan to submit a pull request, you should first
 `fork <https://github.com/mwclient/mwclient#fork-destination-box>`_
-the mwclient repo on GitHub, then clone your own fork:
+the mwclient repo on GitHub, then check out the original repository
+and configure your fork as a remote:
 
 .. code:: bash
 
-    $ git clone git@github.com:MYUSERNAME/mwclient.git
+    $ git clone https://github.com/mwclient/mwclient.git
     $ cd mwclient
+    $ git remote add fork git@github.com:MYUSERNAME/mwclient.git
 
 You can then use pip to do an "editable" install so that your
 edits will be immediately available for (both interactive
@@ -26,6 +28,12 @@ and automated) testing:
 .. code:: bash
 
     $ pip install -e .
+
+Create a new branch for your changes:
+
+.. code:: bash
+
+    $ git checkout -b my-branch
 
 Test suite
 ----------
@@ -91,7 +99,13 @@ on the main master branch to ease merging:
 
 .. code:: bash
 
-    $ git remote add upstream https://github.com/mwclient/mwclient.git
-    $ git rebase upstream master
+    $ git rebase master
 
-Then push your code and open a pull request on GitHub.
+When it is ready, push your branch to your remote:
+
+.. code:: bash
+
+    $ git push -u fork my-branch
+
+Then you can open a pull request on GitHub. You should see a URL to do this
+when you push your branch.
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,9 @@ needs_pytest = set(['pytest', 'test', 'ptr']).intersection(sys.argv)
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
 setup(name='mwclient',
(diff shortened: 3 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+      uses: actions/setup-python@v5.1.0
+      with:
+        python-version: "3.x"
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
```
waldyrious, line 22: Why is this line broken like this? :thinking:

waldyrious, line 22: Why is this line broken like this? :thinking:

AdamWill: I don't know, that's how it is in the guide. I guess we can change it.

AdamWill:
BTW, I got notified of a failed run of the new action for this PR. Looking at it, I *think* it failed because it ran from my fork (AdamWill/mwclient), not from the project (mwclient/mwclient). Of course I configured the pypi project to only allow mwclient/mwclient to publish, not forks :) I'm not sure if this would happen every time anyone pushed a commit to their fork, if we merged this; if so that would be rather annoying and we'd have to find a way to avoid it...

.github/workflows/release.yml
waldyrious: Yeah, if we can't figure out a good reason for that, I'd say we should change it. While you're at it, please change "Github" to "GitHub" in the commit message :)

waldyrious:
Yeah, I think that does happen indeed — every now and then I get notifications from GitHub about actions being disabled on forks I had created months earlier to submit a PR. I'm not sure there is a solution, but it sounds annoying enough that probably people have at least investigated that possibility. Unfortunately that's the extent of my knowledge about this.

AdamWill:
I'm also concerned the `publish-to-testpypi` job might fail every time it's run after the first, between releases, because it looks like it'll keep trying to publish the same version over and over...but it's in the sample config, so...hmm. I guess we can take it out if it does that...

AdamWill:
Ah, yeah, it [looks like I'm right about that](https://github.com/pypa/packaging.python.org/issues/804). Will drop the publish-to-testpypi bit, then.

AdamWill:
OK, there, I think that should be good?

AdamWill:
well, added a comment to setup.py explaining how to do a release.

setup.py
```diff
@@ -10,7 +10,12 @@
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
 setup(name='mwclient',
-      version='0.10.1',  # Use bumpversion to update
+      # MAINTAINERS: to cut a release, ensure CHANGELOG.md is updated,
+      # then run `bump-my-version bump major|minor|patch`
+      # (as appropriate), check the commit looks correct and is tagged
+      # vX.Y.Z, and push. the.github/workflows/release.yml action will
+      # publish to PyPi
```
waldyrious, line 17: This is so close to semantic line-breaking — can we re-wrap it to get it all the way there? 😁

```suggestion
      # MAINTAINERS: to cut a release, ensure CHANGELOG.md is updated,
      # then run `bump-my-version bump major|minor|patch` (as suitable),
      # check the commit looks correct and is tagged vX.Y.Z, and push.
      # The .github/workflows/release.yml action will publish to PyPI
```

waldyrious: Btw, I think this should be documented [here](https://github.com/mwclient/mwclient/blob/master/docs/source/development/index.rst). Either the text should be duplicated there, or this comment should link to the [resulting page](https://mwclient.readthedocs.io/en/latest/development/) (ideally with a section anchor).

AdamWill: Eh, I'm not sure. That feels more like a "contributors' guide" than a "maintainers' guide" to me. I felt like this was more likely to be somewhere a future maintainer might actually look when trying to do an update.

AdamWill: ...but hey, if you insist, how about this? (yes, readthedocs is somehow still getting automatically updated, not sure how, but it is!)

docs/source/development/index.rst
```diff
@@ -91,7 +99,27 @@ on the main master branch to ease merging:
...
+    $ git rebase master
+
+When it is ready, push your branch to your remote:
+
+.. code:: bash
+
+    $ git push -u fork my-branch
+
+Then you can open a pull request on GitHub (you should see a URL to do this
+when you push your branch).
```
waldyrious, line 111:
```suggestion
Then you can open a pull request on GitHub
(you should see a URL to do this when you push your branch).
```

docs/source/development/index.rst
```diff
@@ -91,7 +99,27 @@ on the main master branch to ease merging:
...
+
+Making a release
+----------------
+
+These instructions are for maintainers of the project. To cut a release, ensure
+``CHANGELOG.md`` is updated, then use bump-my-version:
```
waldyrious, line 117:
```suggestion
These instructions are for maintainers of the project.
To cut a release, ensure ``CHANGELOG.md`` is updated, then use
`bump-my-version <https://callowayproject.github.io/bump-my-version/>`_:
```

docs/source/development/index.rst
```diff
@@ -91,7 +99,27 @@ on the main master branch to ease merging:
...
+
+.. code:: bash
+
+    $ pip install bump-my-version
+    $ bump-my-version bump major|minor|patch
 
-Then push your code and open a pull request on GitHub.
+Then check the commit looks correct and is tagged vX.Y.Z, and push. The
+``.github/workflows/release.yml`` action will publish to PyPi.
```
waldyrious, line 125:
```suggestion
``.github/workflows/release.yml`` action will publish to PyPI.
```

setup.py
```diff
@@ -10,7 +10,9 @@
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
 setup(name='mwclient',
-      version='0.10.1',  # Use bumpversion to update
+      # see https://mwclient.readthedocs.io/en/latest/development/#making-a-release
+      # for how to update this and make a release
```
waldyrious, line 14:
```suggestion
      # See https://mwclient.readthedocs.io/en/latest/development/#making-a-release
      # for how to update this field and release a new version.
```

setup.py
waldyrious: I see what you mean. I do agree with the way you did it, though — after all, we want to (1) ensure all processes are well documented, for both contributors and maintainers; and (2) provide a smooth progression path from one-time/occasional contributor to regular contributor to maintainer.

I've left a few minor edit suggestions as inline comments, but other than that, LGTM :)

waldyrious:
Looks great! I've left a few minor edit suggestions as inline comments, but other than that, this should be ready to go :)

waldyrious:
Btw, what a nice coincidence that you became active precisely during the [Wikimedia Hackathon](https://www.mediawiki.org/wiki/Wikimedia_Hackathon_2024), so I was available to review your changes quickly. I also happened to bump into @WolfgangFahl, who's [been active in the project's issues](https://github.com/search?q=repo%3Amwclient%2Fmwclient+involves%3AWolfgangFahl+&type=issues), so overall it was a nice weekend for mwclient! 😁

docs/source/development/index.rst
```diff
@@ -91,7 +99,28 @@ on the main master branch to ease merging:
...
+    $ git rebase master
+
+When it is ready, push your branch to your remote:
+
+.. code:: bash
+
+    $ git push -u fork my-branch
+
+Then you can open a pull request on GitHub. You should see a URL to do this
+when you push your branch.
```
waldyrious, line 111: Just to be sure, was it intentional that you did not break the second sentence into its own line?

waldyrious (approved):
Looks great! Feel free to merge at your convenience.

AdamWill:
actually, just sent one final change: made the conditional for the two remaining steps in the workflow the same, because now we don't have the publish-to-testpypi step there's really no point building a dist for every push. if you can just sign off on that too, I'll merge. thanks! @waldyrious

waldyrious:
That looks good, and makes sense. But just to be sure, did you see the inline comment [above](https://github.com/mwclient/mwclient/pull/331#pullrequestreview-2041321849)?

AdamWill:
Yes, I replied to it and closed it.

docs/source/development/index.rst
waldyrious: Oh, right — sorry, I should've been more explicit in my [first comment](https://github.com/mwclient/mwclient/pull/331#discussion_r1590804140). I am not adamant about semantic line breaks, but I like to encourage them when possible.

waldyrious:
Great, let's go then! LGTM :)

AdamWill:
ugh. So now I see we *did* actually do github releases from 0.7.1 onwards, and don't use the Changelog file any more. I guess I'll have to look into the github release bit of the upstream doc here after all, figure that out, and send a followup.

PR #332, opened 2024-05-06
Document the clientlogin method, mark login as deprecated

AdamWill:
I tried to make this as accurate as possible based on the method docstrings and the upstream docs.


marcfrederick (approved):
Looks good, thank you!

PR #334, opened 2024-05-12
Add a sign/GH release job to the release workflow

AdamWill:
This is the one from
https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ again, with an `if` condition added and multilines rewrapped. As best as I can tell, this should work automagically, we don't need to "sign up for" sigstore or create any certificates or keys; it works by creating an ephemeral signing certificate bound to the identity of whatever GitHub user the workflow runs as (proved by OIDC), and signing with that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,3 +45,36 @@ jobs:
         path: dist/
     - name: Publish distribution to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
+
+  github-release:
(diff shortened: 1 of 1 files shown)


AdamWill:
For now I intend to just edit the github release after this creates it and manually write a changelog in there. For the future we could look at using something like https://pypi.org/project/towncrier/ , which lets you provide snippets to build a changelog as part of each PR/commit, then produce the combined changelog automatically on release.

marcfrederick (approved):

PR #336, opened 2024-06-16
Add Type Annotations to the Codebase

marcfrederick:
Since dropping Python 2 support in #310, we now have the ability to add type annotations to the codebase. I think this could be very helpful for users to discover the API and to prevent bugs.

This PR is intended as a starting point for discussions (@mwclient/maintainers) and adds basic type annotations to all functions. Since we often return responses from the MediaWiki API verbatim, I've chosen to leave the return type as `Any` or `Dict[str, Any]` in many cases. This could be improved, for example, with `TypedDict`s, but I'm not sure we want to take on that maintenance burden.

Here are some open questions I have regarding this:
1. Do we want to maintain types in the first place?
2. Which type-checker should we use? I've chosen mypy for now, but we could, for example, use pyright instead.
3. Do we want to type responses, for example, using `TypedDict`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ dist
 .eggs
 .tox
 .idea/*
+.venv/
 .coverage
 coverage.lcov
--- a/mwclient/__init__.py
+++ b/mwclient/__init__.py
@@ -24,7 +24,7 @@
 """
 
 from mwclient.errors import *  # noqa: F401, F403
-from mwclient.client import Site, __version__  # noqa: F401
+from mwclient.client import Site as Site, __version__ as __version__  # noqa: F401
 import logging
 import warnings
 
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -986,7 +986,7 @@ class Site:
         url: Optional[str] = None,
         filekey: Optional[str] = None,
         comment: Optional[str] = None
-    ) -> Any:
+    ) -> Dict[str, Any]:
         """Upload a file to the site.
 
         Note that one of `file`, `filekey` and `url` must be specified, but not
@@ -1098,7 +1098,7 @@ class Site:
             if not info:
                 info = {}
             if self.handle_api_result(info, kwargs=predata, sleeper=sleeper):
-                response = info.get('upload', {})
+                response = info.get('upload', {})  # type: Dict[str, Any]
                 # Workaround for https://github.com/mwclient/mwclient/issues/211
                 # ----------------------------------------------------------------
                 # Raise an error if the file already exists. This is necessary because
@@ -1120,7 +1120,7 @@ class Site:
         ignorewarnings: bool,
         comment: str,
         text: Optional[str]
-    ) -> Any:
+    ) -> Dict[str, Any]:
         """Upload a file to the site in chunks.
 
         This method is called by `Site.upload` if you are connecting to a newer
@@ -1158,7 +1158,7 @@ class Site:
                 data = self.raw_call('api', params, files={'chunk': chunk})
                 info = json.loads(data)
                 if self.handle_api_result(info, kwargs=params, sleeper=sleeper):
-                    response = info.get('upload', {})
+                    response = info.get('upload', {})  # type: Dict[str, Any]
                     break
 
             offset += chunk.tell()
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -685,7 +685,7 @@ class Site:
         minor: int,
         revision: Optional[int] = None,
         raise_error: bool = True
-    ) -> bool:
+    ) -> Optional[bool]:
         """Check whether the current wiki matches the required version.
 
         Args:
@@ -697,8 +697,8 @@ class Site:
 
         Returns:
             `False` if the version of the current wiki is below the required version, else
-                `True`. If either `raise_error=True` or the site is uninitialized and
-                `raise_error=None` then nothing is returned.
+                `True`. If `raise_error` is `False` and the version is below the required
+                version, `None` is returned.
 
         Raises:
             errors.MediaWikiVersionError: The current wiki is below the required version
@@ -713,9 +713,17 @@ class Site:
         """
         if self.version is None:
             if raise_error is None:
-                return  # type: ignore[unreachable]
-            # FIXME: Replace this with a specific error
-            raise RuntimeError(f'Site {repr(self)} has not yet been initialized')
+                warnings.warn(  # type: ignore[unreachable]
+                    'Passing raise_error=None to require is deprecated and will be '
+                    'removed in a future version. Use raise_error=False instead.',
+                    DeprecationWarning
+                )
+                return None
+            elif raise_error is False:
+                return None
+            else:
+                # FIXME: Replace this with a specific error
+                raise RuntimeError(f'Site {repr(self)} has not yet been initialized')
 
         if revision is None:
             if self.version[:2] >= (major, minor):
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -1,7 +1,10 @@
+import io
 import json
 import logging
 import warnings
 from collections import OrderedDict
+from typing import Optional, Callable, Union, Mapping, Any, MutableMapping, List, Dict, \
+    Tuple, cast, Iterable, BinaryIO, TextIO, Iterator
 
 import requests
 from requests.auth import HTTPBasicAuth, AuthBase
@@ -9,7 +12,8 @@ from requests_oauthlib import OAuth1
 
 import mwclient.errors as errors
 import mwclient.listing as listing
-from mwclient.sleep import Sleepers
+from mwclient.sleep import Sleeper, Sleepers
+from mwclient.types import Namespace
 from mwclient.util import parse_timestamp, read_in_chunks, handle_limit
 
 __version__ = '0.11.0'
@@ -28,53 +32,50 @@ class Site:
         >>> wikia_site = mwclient.Site('vim.wikia.com', path='/')
 
     Args:
-        host (str): The hostname of a MediaWiki instance. Must not include a
-            scheme (e.g. `https://`) - use the `scheme` argument instead.
-        path (str): The instances script path (where the `index.php` and `api.php` scripts
-            are located). Must contain a trailing slash (`/`). Defaults to `/w/`.
-        ext (str): The file extension used by the MediaWiki API scripts. Defaults to
-            `.php`.
-        pool (requests.Session): A preexisting :class:`~requests.Session` to be used when
-            executing API requests.
-        retry_timeout (int): The number of seconds to sleep for each past retry of a
-            failing API request. Defaults to `30`.
-        max_retries (int): The maximum number of retries to perform for failing API
-            requests. Defaults to `25`.
-        wait_callback (Callable): A callback function to be executed for each failing
-            API request.
-        clients_useragent (str): A prefix to be added to the default mwclient user-agent.
-            Should follow the pattern `'{tool_name}/{tool_version} ({contact})'`. Check
-            the `User-Agent policy <https://meta.wikimedia.org/wiki/User-Agent_policy>`_
+        host: The hostname of a MediaWiki instance. Must not include a scheme
+            (e.g. `https://`) - use the `scheme` argument instead.
+        path: The instances script path (where the `index.php` and `api.php` scripts are
+            located). Must contain a trailing slash (`/`). Defaults to `/w/`.
+        ext: The file extension used by the MediaWiki API scripts. Defaults to `.php`.
+        pool: A preexisting :class:`~requests.Session` to be used when executing API
+            requests.
+        retry_timeout: The number of seconds to sleep for each past retry of a failing API
+            request. Defaults to `30`.
+        max_retries: The maximum number of retries to perform for failing API requests.
+            Defaults to `25`.
+        wait_callback: A callback function to be executed for each failing API request.
+        clients_useragent: A prefix to be added to the default mwclient user-agent. Should
+            follow the pattern `'{tool_name}/{tool_version} ({contact})'`. Check the
+            `User-Agent policy <https://meta.wikimedia.org/wiki/User-Agent_policy>`_
             for more information.
-        max_lag (int): A `maxlag` parameter to be used in `index.php` calls. Consult the
+        max_lag: A `maxlag` parameter to be used in `index.php` calls. Consult the
             `documentation <https://www.mediawiki.org/wiki/Manual:Maxlag_parameter>`_ for
             more information. Defaults to `3`.
-        compress (bool): Whether to request and accept gzip compressed API responses.
-            Defaults to `True`.
-        force_login (bool): Whether to require authentication when editing pages. Set to
-            `False` to allow unauthenticated edits. Defaults to `True`.
-        do_init (bool): Whether to automatically initialize the :py:class:`Site` on
+        compress: Whether to request and accept gzip compressed API responses. Defaults to
+             `True`.
+        force_login: Whether to require authentication when editing pages. Set to `False`
+            to allow unauthenticated edits. Defaults to `True`.
+        do_init: Whether to automatically initialize the :py:class:`Site` on
             initialization. When set to `False`, the :py:class:`Site` must be initialized
             manually using the :py:meth:`.site_init` method. Defaults to `True`.
-        httpauth (Union[tuple[basestring, basestring], requests.auth.AuthBase]): An
-            authentication method to be used when making API requests. This can be either
-            an authentication object as provided by the :py:mod:`requests` library, or a
-            tuple in the form `{username, password}`. Usernames and passwords provided as
-            text strings are encoded as UTF-8. If dealing with a server that cannot
-            handle UTF-8, please provide the username and password already encoded with
-            the appropriate encoding.
-        connection_options (Dict[str, Any]): Additional arguments to be passed to the
+        httpauth: An authentication method to be used when making API requests. This can
+            be either an authentication object as provided by the :py:mod:`requests`
+            library, or a tuple in the form `{username, password}`. Usernames and
+            passwords provided as text strings are encoded as UTF-8. If dealing with a
+            server that cannot handle UTF-8, please provide the username and password
+            already encoded with the appropriate encoding.
+        connection_options: Additional arguments to be passed to the
             :py:meth:`requests.Session.request` method when performing API calls. If the
             `timeout` key is empty, a default timeout of 30 seconds is added.
-        consumer_token (str): OAuth1 consumer key for owner-only consumers.
-        consumer_secret (str): OAuth1 consumer secret for owner-only consumers.
-        access_token (str): OAuth1 access key for owner-only consumers.
-        access_secret (str): OAuth1 access secret for owner-only consumers.
-        client_certificate (Union[str, tuple[str, str]]): A client certificate to be added
+        consumer_token: OAuth1 consumer key for owner-only consumers.
+        consumer_secret: OAuth1 consumer secret for owner-only consumers.
+        access_token: OAuth1 access key for owner-only consumers.
+        access_secret: OAuth1 access secret for owner-only consumers.
+        client_certificate: A client certificate to be added
             to the session.
-        custom_headers (Dict[str, str]): A dictionary of custom headers to be added to all
+        custom_headers: A dictionary of custom headers to be added to all
             API requests.
-        scheme (str): The URI scheme to use. This should be either `http` or `https` in
+        scheme: The URI scheme to use. This should be either `http` or `https` in
             most cases. Defaults to `https`.
 
     Raises:
@@ -87,17 +88,41 @@ class Site:
     """
     api_limit = 500
 
-    def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
-                 max_retries=25, wait_callback=lambda *x: None, clients_useragent=None,
-                 max_lag=3, compress=True, force_login=True, do_init=True, httpauth=None,
-                 connection_options=None, consumer_token=None, consumer_secret=None,
-                 access_token=None, access_secret=None, client_certificate=None,
-                 custom_headers=None, scheme='https', reqs=None):
+    def __init__(
+        self,
+        host: str,
+        path: str = '/w/',
+        ext: str = '.php',
+        pool: Optional[requests.Session] = None,
+        retry_timeout: int = 30,
+        max_retries: int = 25,
+        wait_callback: Callable[['Sleeper', int, Optional[Any]], Any] = lambda *x: None,
+        clients_useragent: Optional[str] = None,
+        max_lag: int = 3,
+        compress: bool = True,
+        force_login: bool = True,
+        do_init: bool = True,
+        httpauth: Union[
+            Tuple[Union[str, bytes], Union[str, bytes]],
+            requests.auth.AuthBase,
+            List[Union[str, bytes]],
+            None,
+        ] = None,
+        connection_options: Optional[MutableMapping[str, Any]] = None,
+        consumer_token: Optional[str] = None,
+        consumer_secret: Optional[str] = None,
+        access_token: Optional[str] = None,
+        access_secret: Optional[str] = None,
+        client_certificate: Optional[Union[str, Tuple[str, str]]] = None,
+        custom_headers: Optional[Mapping[str, str]] = None,
+        scheme: str = 'https',
+        reqs: Optional[MutableMapping[str, Any]] = None
+    ) -> None:
         # Setup member variables
         self.host = host
         self.path = path
         self.ext = ext
-        self.credentials = None
+        self.credentials = None  # type: Optional[Tuple[str, str, Optional[str]]]
         self.compress = compress
         self.max_lag = str(max_lag)
         self.force_login = force_login
@@ -136,18 +161,18 @@ class Site:
         self.sleepers = Sleepers(max_retries, retry_timeout, wait_callback)
 
         # Site properties
-        self.blocked = False    # Whether current user is blocked
+        self.blocked = False  # type: Union[Tuple[str, str], bool]  # Is user blocked?
         self.hasmsg = False  # Whether current user has new messages
-        self.groups = []    # Groups current user belongs to
-        self.rights = []    # Rights current user has
-        self.tokens = {}    # Edit tokens of the current user
-        self.version = None
+        self.groups = []  # type: List[str]  # Groups current user is in
+        self.rights = []  # type: List[str]  # Rights current user has
+        self.tokens = {}  # type: Dict[str, str]  # Edit tokens of the current user
+        self.version = None  # type: Union[Tuple[int, ...], Tuple[int, str], None]
 
-        self.namespaces = self.default_namespaces
+        self.namespaces = self.default_namespaces  # type: Dict[int, str]
 
         # Setup connection
         if pool is None:
-            self.connection = requests.Session()
+            self.connection = requests.Session()  # type: requests.Session
             self.connection.auth = auth
             if client_certificate:
                 self.connection.cert = client_certificate
@@ -191,7 +216,7 @@ class Site:
                 if e.args[0] not in {'unknown_action', 'readapidenied'}:
                     raise
 
-    def site_init(self):
+    def site_init(self) -> None:
         """Populates the object with information about the current user and site. This is
         done automatically when creating the object, unless explicitly disabled using the
         `do_init=False` constructor argument."""
@@ -229,7 +254,9 @@ class Site:
         self.initialized = True
 
     @staticmethod
-    def version_tuple_from_generator(string, prefix='MediaWiki '):
+    def version_tuple_from_generator(
+        string: str, prefix: str = 'MediaWiki '
+    ) -> Union[Tuple[int, ...], Tuple[int, str]]:
         """Return a version tuple from a MediaWiki Generator string.
 
         Example:
@@ -237,8 +264,8 @@ class Site:
             (1, 5, 1)
 
         Args:
-            string (str): The MediaWiki Generator string.
-            prefix (str): The expected prefix of the string.
+            string: The MediaWiki Generator string.
+            prefix: The expected prefix of the string.
 
         Returns:
             tuple[int, int, Union[int, str]...]: The version tuple.
@@ -248,7 +275,7 @@ class Site:
 
         version = string[len(prefix):]
 
-        def _split_version(version):
+        def _split_version(version: str) -> Iterator[str]:
             """Split a version string into segments.
 
             Args:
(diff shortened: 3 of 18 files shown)


AdamWill:
I agree with @RheingoldRiver on all points.

marcfrederick:
Great! In that case, I'll remove the draft status from this PR and leave it ready for review. If you have existing projects using mwclient, you should be able to test the annotations by installing from this branch:

```
pip install git+https://github.com/mwclient/mwclient.git@feature/add-types
```

.gitignore
```diff
@@ -8,5 +8,8 @@ dist
 .eggs
 .tox
 .idea/*
+.venv/
+.coverage
+coverage.lcov
```
AdamWill, line 13: this seems to already be there, two lines below.

mwclient/client.py
```diff
@@ -89,17 +88,40 @@ class Site:
...
-                 scheme='https'):
+    def __init__(
+        self,
+        host: str,
+        path: str = '/w/',
+        ext: str = '.php',
+        pool: Optional[requests.Session] = None,
+        retry_timeout: int = 30,
+        max_retries: int = 25,
+        wait_callback: Callable[..., None] = lambda *x: None,
```
AdamWill, line 99: I don't think `...` is correct here. The doc says "If a literal ellipsis ... is given as the argument list, it indicates that a callable with any arbitrary parameter list would be acceptable", but this callable has a specific expected signature. It's ultimately called (in `sleep.py`) as `self.callback(self, self.retries, self.args)`. `self.args` comes from a call to `Sleepers.make` and defaults to `None`, and all the calls to `self.sleepers.make` in `client.py` are just `self.sleepers.make()`, so `self.args` for the individual sleeper will always be `None`. So I think this callback must accept two args, the first being an `int` and the second being `None`. On the other hand, I don't think it has to *return* `None`, because its return value is ignored when it's called, so it could return anything. So...I think this should be `Callable[[int, None], Any]`. Or am I wrong? (this is a pretty odd interface, now I dig into it, I'd be surprised if anyone figured out how to use it...:>)

AdamWill: I guess *technically* it should be something like `Callable[[int, Optional[Any]], Any]`...

mwclient/sleep.py
```diff
@@ -17,24 +19,30 @@ class Sleepers:
...
+        retry_timeout: The time to sleep for each past retry.
+        callback: A callable to be called on each retry.
     """
-    def __init__(self, max_retries, retry_timeout, callback=lambda *x: None):
+
+    def __init__(
+        self,
+        max_retries: int,
+        retry_timeout: int,
+        callback: Callable[..., None] = lambda *x: None
```
AdamWill, line 35: see earlier note, I don't think `...` or `None` is right here.

mwclient/sleep.py
```diff
@@ -48,25 +56,32 @@ class Sleeper:
...
+        callback: A callable to be called on each retry.
     """
-    def __init__(self, args, max_retries, retry_timeout, callback):
+
+    def __init__(
+        self,
+        args: Any,
+        max_retries: int,
+        retry_timeout: int,
+        callback: Callable[..., None]
```
AdamWill, line 76: see earlier note, I don't think `...` or `None` is right here.

AdamWill (changes requested):
Oof, I started reviewing this then realized it's gonna need, like, hours, so I shouldn't be doing it on a work day. Will get back to it later. For now, here's a couple notes.

mwclient/client.py
marcfrederick: Yes, you're correct. I had just taken the signature from the old docstring. Actually, I think the correct type is `Callable[['Sleeper', int, Optional[Any]], Any]` since the `sleep.Sleeper` is passed as the first argument:

```python
class Sleeper:
  # ...
  def sleep(self, min_time=0):
    # ...
    self.callback(self, self.retries, self.args)
```

I've pushed an updated version.

mwclient/__init__.py
```diff
@@ -24,7 +24,7 @@
 """
 
 from mwclient.errors import *  # noqa: F401, F403
-from mwclient.client import Site, __version__  # noqa: F401
+from mwclient.client import Site as Site, __version__ as __version__  # noqa: F401
```
AdamWill, line 27: why?

marcfrederick: It's required by mypy due to `--no-implicit-reexport`, which is part of `--strict`. Without this we get warning such as the following:

```
test/test_client.py:90: error: Module "mwclient" does not explicitly export attribute "Site"  [attr-defined]
test/test_client.py:129: error: Module "mwclient" does not explicitly export attribute "__version__"  [attr-defined]
...
```

The alternative is to disable `--no-implicit-reexport` in the mypy config.

mwclient/client.py
```diff
@@ -137,18 +161,18 @@ def __init__(self, host, path='/w/', ext='.php', pool=None, retry_timeout=30,
...
+        self.blocked = False  # type: Union[Tuple[str, str], bool]  # Is user blocked?
         self.hasmsg = False  # Whether current user has new messages
-        self.groups = []    # Groups current user belongs to
-        self.rights = []    # Rights current user has
-        self.tokens = {}    # Edit tokens of the current user
-        self.version = None
+        self.groups = []  # type: List[str]  # Groups current user is in
+        self.rights = []  # type: List[str]  # Rights current user has
+        self.tokens = {}  # type: Dict[str, str]  # Edit tokens of the current user
+        self.version = None  # type: Union[Tuple[int, ...], Tuple[int, str], None]
```
AdamWill, line 169: I'm not sure about this one. After playing around with the code that produces this for a bit, the rules seem to be: it will always a tuple, it will be *at least* two items long, and the first two items will be ints. After that, any given item may be a string or an int. E.g. `mwclient.Site.version_tuple_from_generator("MediaWiki 1.5.2a.1.3b.1")` gives `(1, 5, 2, 'a', 1, 3, 'b', 1)`. So I think what we want here is something like `Optional[Tuple[Union[int, str], ...]]`, if that's valid?

mwclient/client.py
```diff
@@ -262,9 +288,9 @@ def split_num(s):
                     break
                 i += 1
             if s[i:]:
-                return (int(s[:i]), s[i:], )
+                return (int(s[:i]), s[i:],)
             else:
-                return (int(s[:i]), )
+                return (int(s[:i]),)
 
         version_tuple = sum((split_num(s) for s in version), ())
```
AdamWill, line 295: I just wasted fifteen minutes figuring out what the hell this does...it flattens the sequence of tuples into one tuple...probably would be more idiomatic to do `tuple(itertools.chain(split_num(s) for s in version)))`...

mwclient/client.py
marcfrederick: Yeah seems like you're right, that function is _weird_. I've added a new type `VersionTuple = Tuple[Union[int, str], ...]` to the `types.py` that is used in both locations.

mwclient/client.py
```diff
@@ -230,16 +254,18 @@ def site_init(self):
         self.initialized = True
 
     @staticmethod
-    def version_tuple_from_generator(string, prefix='MediaWiki '):
+    def version_tuple_from_generator(
+        string: str, prefix: str = 'MediaWiki '
+    ) -> Union[Tuple[int, ...], Tuple[int, str]]:
```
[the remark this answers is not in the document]
marcfrederick: Fixed in f2d9819

mwclient/client.py
```diff
@@ -433,7 +469,16 @@ def _query_string(*args, **kwargs):
         ]
         return OrderedDict(qs1 + qs2)
 
-    def raw_call(self, script, data, files=None, retry_on_error=True, http_method='POST'):
+    def raw_call(
+        self,
+        script: str,
+        data: Mapping[str, Any],
+        files: Optional[
+            Mapping[str, Union[io.BytesIO, Tuple[str, io.BufferedReader]]]
```
AdamWill, line 477: I think this is a bit too restrictive in various ways, see https://requests.readthedocs.io/en/latest/user/quickstart/#post-a-multipart-encoded-file . It seems like a `Tuple[str, str]` is also valid, and three- or four-item tuples are also possible (to specify content type and headers). And, uh, `files` can also apparently be a list of tuples - https://requests.readthedocs.io/en/latest/user/advanced/#post-multiple-multipart-encoded-files . Whee, isn't typing fun

mwclient/client.py
```diff
@@ -521,7 +568,7 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
                 if not retry_on_error:
                     raise
                 log.warning('Connection error. Retrying in a moment.')
-                toraise = err
+                toraise = err  # type: ignore[assignment]
```
AdamWill, line 571: could we make this `Union[BaseException, str]`, since it has to be something that can be raised or the string `"stream"`?

mwclient/client.py
```diff
@@ -531,20 +578,26 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
                 if toraise == "stream":
                     stream.raise_for_status()
                 elif toraise:
-                    raise toraise
+                    raise toraise  # type: ignore[misc]
```
AdamWill, line 581: same as above

mwclient/client.py
```diff
@@ -628,7 +692,7 @@ def require(self, major, minor, revision=None, raise_error=True):
         """
         if self.version is None:
             if raise_error is None:
-                return
+                return  # type: ignore[unreachable]
```
AdamWill, line 695: This is...weird. It seems like this should be reachable and the check should be `if not raise_error` rather than `if raise_error is None`. This logic seems to first appear in 465d4b0 (the early commit history is weird as Dan was messily syncing from an SVN repo maintained by Bryan, it seems), where it behaves more or less as it currently does, but there is no explicit type hint for `raise_error` except that it defaults to `True`. There's no explanation for why it uses `if raise_error is None` here, but `if raise_error` on the other branch.

In any case, the way it is right now, we would raise an error here with `raise_error=False`, which seems pretty weird. I feel like we should correct this so you can reach this outcome without violating the expected type of `raise_error`, and allow this function to return a `bool` or `None`?

mwclient/client.py
```diff
@@ -676,11 +744,17 @@ def email(self, user, text, subject, cc=False):
...
+            raise errors.EmailError(*e)  # type: ignore[misc]
 
         return info
 
-    def login(self, username=None, password=None, cookies=None, domain=None):
+    def login(
+        self,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        cookies: Optional[Mapping[str, str]] = None,
```
AdamWill, line 755: fun note: requests allows this to be an `http.cookiejar.CookieJar` as well as a dict, but we effectively don't because we do `self.connection.cookies.update(cookies)` and `CookieJar` instances do not have an `update` method.

mwclient/client.py
```diff
@@ -754,7 +828,9 @@ def login(self, username=None, password=None, cookies=None, domain=None):
 
         self.site_init()
 
-    def clientlogin(self, cookies=None, **kwargs):
+    def clientlogin(
+        self, cookies: Optional[Mapping[str, str]] = None, **kwargs: Any
```
AdamWill, line 832: see above re `cookies` can be `CookieJar`

mwclient/client.py
```diff
@@ -941,10 +1028,15 @@ def upload(self, file=None, filename=None, description='', ignore=False,
             if not hasattr(file, 'read'):
                 file = open(file, 'rb')
 
+            # Narrowing the type of file from Union[str, io.BufferedReader] to just
```
AdamWill, line 1031: it was never `Union[str, io.BufferedReader]`, though?

mwclient/client.py
```diff
@@ -1001,16 +1092,26 @@ def upload(self, file=None, filename=None, description='', ignore=False,
             file.close()
         return response
 
-    def chunk_upload(self, file, filename, ignorewarnings, comment, text):
+    def chunk_upload(
+        self,
+        file: io.BufferedReader,
```
AdamWill, line 1097: why are we using `io.BufferedReader` here but `Union[BinaryIO, TextIO]` for essentially the same thing in `upload()`? It seems like maybe you got kinda caught between two options and it's inconsistent? I don't know which is better, but it seems like we should consistently use one or the other when we mean "the thing you get from opening a file"?

mwclient/client.py
```diff
@@ -877,25 +955,34 @@ def get_token(self, type, force=False, title=None):
...
+        self,
+        file: Union[str, BinaryIO, TextIO, None] = None,
+        filename: Optional[str] = None,
+        description: str = '',
+        ignore: bool = False,
+        file_size: Optional[int] = None,
+        url: Optional[str] = None,
+        filekey: Optional[str] = None,
+        comment: Optional[str] = None
+    ) -> Any:
```
AdamWill, line 968: I think the return value will always be an API response, we can use `Dict[str, Any]` here like we do for `api()`.

mwclient/client.py
```diff
@@ -1257,21 +1457,44 @@ def blocks(self, start=None, end=None, dir='older', ids=None, users=None, limit=
...
 
-    def deletedrevisions(self, start=None, end=None, dir='older', namespace=None,
-                         limit=None, prop='user|comment', max_items=None,
-                         api_chunk_size=None):
+    def deletedrevisions(
+        self,
+        start: Optional[str] = None,
+        end: Optional[str] = None,
+        dir: str = 'older',
+        namespace: Optional[int] = None,
```
AdamWill, line 1465: should this be `Namespace`?

mwclient/client.py
```diff
@@ -1349,7 +1623,9 @@ def recentchanges(self, start=None, end=None, dir='older', namespace=None,
         return listing.List(self, 'recentchanges', 'rc', max_items=max_items,
                             api_chunk_size=api_chunk_size, **kwargs)
 
-    def revisions(self, revids, prop='ids|timestamp|flags|comment|user'):
+    def revisions(
+        self, revids: List[int], prop: str = 'ids|timestamp|flags|comment|user'
```
AdamWill, line 1627: I think `revids` could be ints or strs, since we actually force them to str in the function body anyway.

mwclient/client.py
```diff
@@ -1459,16 +1769,18 @@ def watchlist(self, allrev=False, start=None, end=None, namespace=None, dir='old
         return listing.List(self, 'watchlist', 'wl', max_items=max_items,
                             api_chunk_size=api_chunk_size, **kwargs)
 
-    def expandtemplates(self, text, title=None, generatexml=False):
+    def expandtemplates(
+        self, text: str, title: Optional[str] = None, generatexml: bool = False
+    ) -> Any:
```
AdamWill, line 1774: I think we can make this stricter - the return value should always be a string or a 2-tuple of strings. (this function definitely needs updating - the format it's using is deprecated - and probably needs more error checking - it's pretty gung-ho about assuming it gets an "OK" response!)

mwclient/client.py
```diff
@@ -1485,14 +1797,14 @@ def expandtemplates(self, text, title=None, generatexml=False):
         else:
             return result['expandtemplates']['*']
 
-    def ask(self, query, title=None):
+    def ask(self, query: str, title: Optional[str] = None) -> Iterable[Dict[str, Any]]:
```
AdamWill, line 1800: Shouldn't we be using `Generator` here, not `Iterable`?

mwclient/client.py
```diff
@@ -1507,16 +1819,16 @@ def ask(self, query, title=None):
             >>>         print(title)
             >>>         print(data)
         """
-        kwargs = {}
+        kwargs = {}  # type: Dict[str, Any]
```
AdamWill, line 1822: since the only thing we ever put in `kwargs` is the title, this could just be `Dict[str, str]`, right?

Actually, er, isn't that code wrong and the check should be `if title is not None:`?

AdamWill:
whew, okay, made it through `client.py` :P Gonna take a break now.

mwclient/client.py
marcfrederick: Yeah, the actual type (according to `types-requests` is something like the following:

```python
files: Mapping[str, SupportsRead[str | bytes] | str | bytes | tuple[str | None, SupportsRead[str | bytes] | str | bytes] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, Mapping[str, str]]] | Iterable[tuple[str, SupportsRead[str | bytes] | str | bytes | tuple[str | None, SupportsRead[str | bytes] | str | bytes] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str] | tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, Mapping[str, str]]]] | None = None
```

I think we just have to come up with a reasonable subset here to be honest 😅

mwclient/client.py
marcfrederick: Technically it should be `Optional[Union[requests.RequestException, Literal['stream']]]` I guess, which would make the if/else type-check cleanly, but `Litaral` requires Python 3.8+. For now we need to go with something like this I guess:

```diff
diff --git a/mwclient/client.py b/mwclient/client.py
index a1e0cde..5593a32 100644
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -527,7 +527,7 @@ class Site:
         url = f'{scheme}://{host}{self.path}{script}{self.ext}'
 
         while True:
-            toraise = None
+            toraise = None  # type: Optional[Union[requests.RequestException, str]]
             wait_time = 0
             args = {'files': files, 'headers': headers}  # type: Dict[str, Any]
             for k, v in self.requests.items():
@@ -568,7 +568,7 @@ class Site:
                 if not retry_on_error:
                     raise
                 log.warning('Connection error. Retrying in a moment.')
-                toraise = err  # type: ignore[assignment]
+                toraise = err
                 # proceed to the sleep
 
             # all retry paths come here
@@ -577,8 +577,8 @@ class Site:
             except errors.MaximumRetriesExceeded:
                 if toraise == "stream":
                     stream.raise_for_status()
-                elif toraise:
-                    raise toraise  # type: ignore[misc]
+                elif toraise and isinstance(toraise, Exception):
+                    raise toraise
                 else:
                     raise
```

mwclient/client.py
marcfrederick: We could use `Generator[Dict[str, Any], None, None]`, but according to the docs (https://docs.python.org/3/library/typing.html#typing.Generator) it's fine to use `Iterator[Dict[str, Any]]` instead. I think this is a bit more readable as it is, but am also fine with changing it.

AdamWill: oh, sure, if the docs say it's fine I guess it's fine.

mwclient/client.py
marcfrederick: Yeah, that makes sense. However, I would like to avoid mixing typing with functional changes. Let's create a separate PR either changing this condition or removing the parameter altogether. The API:Ask doesn't even have a `title` parameter from what I can see, neither in the initial nor the current implementation 😅:

- Initial implementation: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3052/files#diff-40638df1e5264aa7030d823f1f90c37081adfd76dd8da007f76d7bbc194987b1L50-L57
- Current implementation: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/d3ef92592b4366d7a4c8cfc1c858956f4f5bfc60/src/MediaWiki/Api/Ask.php#L55-L67
- Docs: https://www.semantic-mediawiki.org/wiki/Help:API:ask

mwclient/client.py
marcfrederick: Yep, I've updated this.

mwclient/client.py
marcfrederick: Changed.

mwclient/client.py
marcfrederick: According to the MediaWiki documentation, the only allowed values for this are integers (https://www.mediawiki.org/wiki/API:Deletedrevs#query+deletedrevs:drnamespace). Most other APIs also allow `*` (all namespaces). While we could technically allow `str`, since the function works fine with strings, using int enforces correct usage (i.e., no `*`).

mwclient/client.py
marcfrederick: We can, but it requires some casts:

```diff
diff --git a/mwclient/client.py b/mwclient/client.py
index abb3f5e..7e65dc3 100644
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -965,7 +965,7 @@ class Site:
         url: Optional[str] = None,
         filekey: Optional[str] = None,
         comment: Optional[str] = None
-    ) -> Any:
+    ) -> Dict[str, Any]:
         """Upload a file to the site.
 
         Note that one of `file`, `filekey` and `url` must be specified, but not
@@ -1090,7 +1090,7 @@ class Site:
 
         if file is not None:
             file.close()
-        return response
+        return cast(Dict[str, Any], response)
 
     def chunk_upload(
         self,
@@ -1099,7 +1099,7 @@ class Site:
         ignorewarnings: bool,
         comment: str,
         text: Optional[str]
-    ) -> Any:
+    ) -> Dict[str, Any]:
         """Upload a file to the site in chunks.
 
         This method is called by `Site.upload` if you are connecting to a newer
@@ -1154,7 +1154,7 @@ class Site:
                 # get the parameters we need to continue, so we should return
                 # the response now.
                 file.close()
-                return response
+                return cast(Dict[str, Any], response)
 
         del params['action']
         del params['stash']
```

mwclient/client.py
marcfrederick: I think personally I would keep the types as is, deprecate the current behavior, and introduce a new `raise_error is False` condition. I would like to avoid complicating the types, as it would prevent simply conditions such as the following from type-checking.

```python
if site.require(1, 27, raise_error=False):
    ...
```

Instead we could do something like the following, keeping the old behavior but having it fail type-checks and logging a `DeprecationWarning`.


```diff
diff --git a/mwclient/client.py b/mwclient/client.py
index 8f192ff..92e8be8 100644
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -692,7 +692,14 @@ class Site:
         """
         if self.version is None:
             if raise_error is None:
-                return  # type: ignore[unreachable]
+                warnings.warn(  # type: ignore[unreachable]
+                    'Passing raise_error=None to require is deprecated and will be '
+                    'removed in a future version. Use raise_error=False instead.',
+                    DeprecationWarning
+                )
+                return
+            elif raise_error is False:
+                return False
             # FIXME: Replace this with a specific error
             raise RuntimeError(f'Site {repr(self)} has not yet been initialized')
```

mwclient/client.py
marcfrederick: Alternatively we could require typing-extensions on python versions <3.8:

```diff
diff --git a/mwclient/client.py b/mwclient/client.py
index 18e3b09..991759f 100644
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -1,11 +1,17 @@
 import io
 import json
 import logging
+import sys
 import warnings
 from collections import OrderedDict
 from typing import Optional, Callable, Union, Mapping, Any, MutableMapping, List, Dict, \
     Tuple, cast, Iterable, BinaryIO, TextIO
 
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal  # noqa: F401
+else:
+    from typing import Literal  # noqa: F401
+
 import requests
 from requests.auth import HTTPBasicAuth, AuthBase
 from requests_oauthlib import OAuth1
@@ -527,7 +533,7 @@ class Site:
         url = f'{scheme}://{host}{self.path}{script}{self.ext}'
 
         while True:
-            toraise = None
+            toraise = None  # type: Union[Literal['stream'], requests.exceptions.RequestException, None]  # noqa: E501
             wait_time = 0
             args = {'files': files, 'headers': headers}  # type: Dict[str, Any]
             for k, v in self.requests.items():
@@ -568,7 +574,7 @@ class Site:
                 if not retry_on_error:
                     raise
                 log.warning('Connection error. Retrying in a moment.')
-                toraise = err  # type: ignore[assignment]
+                toraise = err
                 # proceed to the sleep
 
             # all retry paths come here
@@ -578,7 +584,7 @@ class Site:
                 if toraise == "stream":
                     stream.raise_for_status()
                 elif toraise:
-                    raise toraise  # type: ignore[misc]
+                    raise toraise
                 else:
                     raise
 
diff --git a/setup.py b/setup.py
index 9420efe..a92fdbd 100644
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,10 @@ setup(name='mwclient',
       url='https://github.com/mwclient/mwclient',
       license='MIT',
       packages=['mwclient'],
-      install_requires=['requests-oauthlib'],
+      install_requires=[
+          'requests-oauthlib',
+          'typing-extensions; python_version < "3.8"',
+      ],
       extras_require={
           'testing': ['pytest', 'pytest-cov',
                       'responses>=0.3.0', 'responses!=0.6.0', 'setuptools'],
```

mwclient/client.py
AdamWill: But that doesn't address the weird situation that calling it with `raise_error=False` before the site is initialized will raise an error, the thing you specifically told it not to do...

marcfrederick: Alright, then let's change the return type to `Optional[bool]` after all. I was wrong about this preventing calls such as `if site.require(1, 27, raise_error=False)`, as `None` is falsy. I'll still add the `DeprecationWarning` to the `raise_error is None` case, as that is just straight up _weird_.

mwclient/client.py
marcfrederick: Fixed in ab9d64c

mwclient/client.py
marcfrederick: The code has been removed in #359 🚀.

test/integration.py
```diff
@@ -33,46 +34,46 @@ def site(request):
     container = f"mwclient-{tag}"
     # create the container, using upstream's official image. see
     # https://hub.docker.com/_/mediawiki
-    args = (_CONTAINER_RUNTIME, "run", "--name", container, "-p", f"{port}:80",
-            "-d", f"docker.io/library/mediawiki:{tag}")
+    args = [_CONTAINER_RUNTIME, "run", "--name", container, "-p", f"{port}:80",
```
AdamWill, line 37: why change all of these? I habitually use tuples for cases where the thing doesn't need to be mutable, on the theory it's possibly slightly faster or more memory-efficient or something. does it cause issues for typing? typeshed doesn't seem to expect these to be lists specifically, just sequences...

test/test_client.py
```diff
@@ -731,47 +733,6 @@ def test_revisions(self):
         assert revisions[1]['revid'] == 689816909
 
 
-class TestVersionTupleFromGenerator:
```
AdamWill, line 734: er, I don't think we want to completely remove a test case? I assume this is a mistake from rebasing on your version parsing improvement PR...

mwclient/client.py
AdamWill: yeah, fair enough.

mwclient/client.py
AdamWill: I think I prefer the former, WDYT?

mwclient/client.py
AdamWill: OK. I don't mind which way around you want to do it (fix/drop it before this, or after).

AdamWill:
a general note - I noticed you quote class names throughout the PR. I can't find an authoritative reference on whether this is the best choice or not. it seems like it's only *required* if the class being referenced might not be defined at the time the type is parsed (e.g. if something in a class can return an instance of that class). do you have any notes on this?

AdamWill:
ping @marcfrederick , will you be able to get to rebasing this soon? it'd be awesome to get it merged before we do anything else substantial. thanks!

test/integration.py
marcfrederick: The issue is not with the signature of `subprocess.run`, but rather with the redefinition of the `args` variable using tuples of different lengths:

```
test/integration.py:43: error: Incompatible types in assignment (expression has type tuple[str, str, ... <17 more items>], variable has type tuple[str, str, str, str, str, str, str, str])  [assignment]
test/integration.py:49: error: Incompatible types in assignment (expression has type tuple[str, str, ... <9 more items>], variable has type tuple[str, str, str, str, str, str, str, str])  [assignment]
test/integration.py:54: error: Incompatible types in assignment (expression has type tuple[str, str, ... <12 more items>], variable has type tuple[str, str, str, str, str, str, str, str])  [assignment]
...
```

The arity of a tuple is part of its type (e.g., `tuple[str, str]` is a different type from `tuple[str, str, str]`), so essentially, the type of the `args` variable keeps changing. From mypy’s perspective, this is equivalent to the following, which would clearly fail type-checking:

```python
foo = 1
foo = "bar"
foo = True
```

To work around this would  either
1. have to set `allow_redefinition = true` in the `tools.mypy` section of the `pyproject.toml` (that would also allow the stuff like my previous example though)
2. use different variables (i.e. `args1`, `args2`, …)
3. use lists instead as I did here. The length of a list is not part of its type, so everything works as it should

You're right that tuples would (probably) be slightly more efficient than lists here, but in the context here I think it doesn't really matter.

test/test_client.py
marcfrederick: Yep, thank you for catching that.

marcfrederick:
> a general note - I noticed you quote class names throughout the PR. I can't find an authoritative reference on whether this is the best choice or not. it seems like it's only _required_ if the class being referenced might not be defined at the time the type is parsed (e.g. if something in a class can return an instance of that class). do you have any notes on this?

Yes, that's one of the reasons for using quotes. From a typing and mypy perspective, most of the quoted references could be removed. However, this leads to runtime issues, such as:

```
AttributeError: partially initialized module 'mwclient' has no attribute 'listing' (most likely due to a circular import)
```

I think fixing this would require a whole lot of restructuring, e.g. at the moment we have a circular import between `listing` and `page`. This does not seem to matter much in practice, until I unquote the types. When quoting the class-names, evaluation is delayed and everything works fine. There might be a better way to do this, but I haven't come up with anything as of now.

mwclient/client.py
marcfrederick: Yep, let's go with that for now. Fixed in 1baa435.

mwclient/client.py
marcfrederick: See 1baa435.

test/integration.py
AdamWill: I see, that makes sense. Thanks!

marcfrederick:
> ping @marcfrederick , will you be able to get to rebasing this soon? it'd be awesome to get it merged before we do anything else substantial. thanks!

Sorry, I was on vacation for the past two weeks. I've rebased this on top of `upstream/master`, and made some small fixups. I think with that this is more or less ready.

AdamWill (approved):

AdamWill:
Thanks again! I merged #362 , so this needs one more rebase, then I think we can land it! Thanks for all the work.

marcfrederick:
Yep, will do. Should we squash this PR before merging?

AdamWill:
yeah, I think a squash is appropriate. I'm fine with doing a squash-and-merge , but if you want to do a manual one for any reason, go ahead. I'll give it a few hours.

PR #337, opened 2024-06-29
Made 429 Too Many Request HTTP error in `raw_call()` of `client.py` retryable

murolem:
This prevents the program from straight-up failing, instead resorting to retries if they are enabled.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -494,7 +494,10 @@ class Site:
                     # fall through to the sleep
                 elif stream.status_code == 200:
                     return stream.text
-                elif (stream.status_code < 500 or stream.status_code > 599) and not stream.status_code == 429:  # 429 is Too Many Request - this is retryable:
+                elif (
+                    (stream.status_code < 500 or stream.status_code > 599)
+                    and stream.status_code != 429  # 429 Too Many Requests is retryable
+                ):
                     stream.raise_for_status()
                 else:
                     if not retry_on_error:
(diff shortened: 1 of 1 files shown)


RheingoldRiver:
This lgtm - fixes a persistent problem with the lib when connecting to wikis that have CF ratelimites

AdamWill:
LGTM too, but it fails tests because the line is too long, our max is 90 chars. @murolem can you reformat? Thanks!

marcfrederick:
I think `stream.status_code != 429` would be easier to read here than `not stream.status_code == 429`. Apart from that ([and the line length](https://github.com/mwclient/mwclient/actions/runs/9723930239/job/27573012035?pr=337)), this looks good to me.

mwclient/client.py
```diff
@@ -494,7 +494,7 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
                     # fall through to the sleep
                 elif stream.status_code == 200:
                     return stream.text
-                elif stream.status_code < 500 or stream.status_code > 599:
+                elif (stream.status_code < 500 or stream.status_code > 599) and not stream.status_code == 429:  # 429 is Too Many Request - this is retryable:
```
marcfrederick, line 497:
```suggestion
                elif (
                    (stream.status_code < 500 or stream.status_code > 599)
                    and stream.status_code != 429  # 429 Too Many Requests is retryable
                ):
```

marcfrederick (changes requested):

marcfrederick (approved):
Thank you @murolem! LGTM now.

PR #338, opened 2024-07-03
don't use bytes in repr()

lilydjwg:
bytes make the output weird and hard to recognize for non-ASCII characters.


AdamWill:
LGTM, @RheingoldRiver acked it in #284 so let's go ahead and merge. Thanks!

Issue #304, opened 2023-08-15
Support asynchronous upload


Gowee:
Hi!

There are some persistent issues with Wikimedia's servers ([T335243](https://phabricator.wikimedia.org/T335243), [T340901](https://phabricator.wikimedia.org/T340901), [T340917](https://phabricator.wikimedia.org/T340901)), which effectively prevents large files to be uploaded.

It appears that the problem can be mitigated by toggling the async flag when calling [API:Upload](https://www.mediawiki.org/wiki/API:Upload), as noted [here](https://commons.wikimedia.org/wiki/User:MidleadingBot#Guide_to_recover_failed_uploads_from_upload_stash). I have conducted tests utilizing the async option of pywikibot, resulting in notable improvements in upload stability when dealing with files of several gigabytes in size.

I suggest that we consider incorporating this option into the mwclient library. This could lead to a more efficient and smoother uploading process for users who handle large files via the library.

PR #339, opened 2024-07-06
Allow the user to set the "async" parameter when uploading files

st4r1ight:
Per #304, this PR allows for setting the "async" parameter on [API:Upload](https://mediawiki.org/wiki/API:Upload) when calling `Site.upload()`. Since the parameter only affects server-side behaviour, this is a pretty small change. It adds the "asynchronous" argument ("async" is a reserved keyword in newer versions of Python), set to False by default for backwards compatibility. When set to True, `{async: 'true'}` is added to `predata`. I added a snippet to the documentation as well.


mwclient/client.py
```diff
@@ -892,6 +892,8 @@ def upload(self, file=None, filename=None, description='', ignore=False,
                            stashed temporarily.
             comment (str): Upload comment. Also used as the initial page text
                            for new files if `description` is not specified.
+            asynchronous (str): Whether the server should make large file
```
marcfrederick, line 895: The type of the `asynchronous` argument should be `bool` instead of `str`.
```suggestion
            asynchronous (bool): Whether the server should make large file
```

marcfrederick (changes requested):
~Hey @st4r1ight, thanks for your contribution! This looks good, and from my local testing, it seems to work (as in the upload still works, though it's not really clear to me from the MW docs what that flag does).~

Actually, I don't think this will work like that. I previously tested with a default MW config, meaning async uploads were actually disabled. After setting `$wgEnableAsyncUploads = true;` in my `LocalSettings.php`, the upload fails with the following error:

> The parameters `async` and `file` cannot be combined. If you want asynchronous processing of your uploaded file, first upload it to stash (using the `stash` parameter) and then publish the stashed file asynchronously (using `filekey` and `async`).

AdamWill:
Can you please include tests for this? I'm currently trying to get test coverage to 100% and would prefer not to merge any new features without tests. Thanks!

AdamWill:
also, you could probably just squash all the changes into one commit.

st4r1ight:
The async feature is not very well documented, but having run some tests in the API sandbox of my own test wiki I believe the flow goes like this:

1. Upload a file using the `stash=true` parameter to upload the file without actually creating a page for it.
2. Take the "filekey" parameter from the response.
3. Call the upload API again using asynchronous=true and the "filekey" from step 2.

The asynchronous bit is the creation of the File page in the database, rather than the actual upload. 

What I've done is have `upload()` throw an error if you try to use `asynchronous` without `filekey` and added a `stash` parameter so that you can get the `filekey` in the first place. People who want to use the `asynchronous` parameter will have to call `upload()` twice, once to stash and once to asynchronously upload. It's a little clunky but I didn't want to do anything more drastic and substantially change the interface for uploading files.

st4r1ight:
I just saw the previous comment about testing which I will do as well.

st4r1ight:
There's now a test to make sure uploading fails if you use `asynchronous` without `filekey`. I couldn't think of any other tests but if you want others I'm happy to write them. Otherwise I'll squash everything into a single commit and prepare to merge.

AdamWill:
I would also expect a test that asserts this works as expected, e.g. by mocking out `handle_api_result` and checking it is called with the expected args. Of course the code being tested is very simple and will certainly work, but the point of having a test is it functions as a kind of declaration that this is intended functionality - say somebody decided to refactor `upload()` entirely and messed it up somehow so the args didn't reach `handle_api_result`, the test will fail and tell them they got it wrong.

BTW, a question I have on looking at this - how do I get the filekey when stashing a file? Presumably it's part of the JSON response that `upload()` returns? Could we maybe at least have a little comment telling people how to get it from that response, or something?

st4r1ight:
There is now:

1. A test that mocks raw_call and checks that 'asynchronous=true' got passed through correctly
2. An explanation of how to do async uploads in the docstring for upload()
3. An example of async uploads in that docstring.

If everything's alright (I will admit to being a novice with Python unit tests) I'll squash the commits for merging.

test/test_client.py
```diff
@@ -862,6 +862,25 @@ def test_upload_missing_upload_permission(self):
...
+        with pytest.raises(TypeError):
+            self.site.upload(filename='Test', file=BytesIO(b'test'), asynchronous=True)
+
+    def test_async_args(self):
+        self.configure()
+
+        self.site.upload(filename='Test', filekey='abc', asynchronous=True)
+
+        args, kwargs = self.raw_call.call_args
+        logging.debug("ASYNC ARGS: ", args)
```
marcfrederick, line 877: This `logging.debug` call seems to be a leftover from development. I think we should remove it.


```suggestion
```

mwclient/client.py
```diff
@@ -1039,12 +1044,24 @@ def upload(
...
+            stash: If set, the file will be stashed instead of uploaded right away.
 
         Example:
 
             >>> client.upload(open('somefile', 'rb'), filename='somefile.jpg',
                               description='Some description')
 
+        Async Example:
+            >>> response = client.upload(open('somefile','rb'),
+                                        filename='somefile.jpg'
```
marcfrederick, line 1058: There's a missing comma in this example:


```suggestion
                                        filename='somefile.jpg',
```

marcfrederick (commented):
Hey @st4r1ight, thanks for the great work! Sorry for the delay on the review. I found two minor issues but otherwise this looks good to merge.

AdamWill:
Agreed with @marcfrederick . Thanks for all your work on this.

st4r1ight:
Just one more thing. I haven't done a PR where you squash into one comment before committing. Should I fix these, then squash them into the current single commit and force push?  Or just fix them and leave them as their own comments on top of the single commit.

AdamWill:
I'd prefer a single squashed commit.

mwclient/client.py
```diff
@@ -1039,12 +1044,24 @@ def upload(
...
+            stash: If set, the file will be stashed instead of uploaded right away.
 
         Example:
 
             >>> client.upload(open('somefile', 'rb'), filename='somefile.jpg',
                               description='Some description')
 
+        Async Example:
+            >>> response = client.upload(open('somefile','rb'),
+                                        filename='somefile.jpg'i,
```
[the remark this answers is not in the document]
st4r1ight: Can you tell that I just started using a modal editor? It's fixed now.

marcfrederick (approved):

moenk:
Using the mwclient 0.11.0 from PyPI not having this option?

AdamWill:
Sorry, we haven't cut a release for some time :( All the current maintainers do this in our spare time and we don't have a lot of it. We also have kind of a backlog of stuff we want to put in releases which slows things down. If I can find an hour or two I'll look at what we've merged since 0.11.0 and see if maybe we want to just go ahead and cut a 0.12.0.

PR #345, opened 2024-08-07
Remove checks for/requirements on writeapi

andrewbogott:
writeapi is being removed

Bug: T294397


examples/basic_edit.py
```diff
@@ -53,12 +52,6 @@
 history = list(image.imagehistory())
 print('History:', history)
 
-if site.writeapi:
```
lucaswerkmeister, line 56: Probably better to keep this code and only remove the condition instead?

lucaswerkmeister (approved):

AdamWill:
For the record: the upstream comment notes "Deprecated since MW 1.32", and 1.32 went out of support in 2020. So I think it should be fine to remove this now.

AdamWill:
This LGTM. @mwclient/maintainers , anyone else want to take a look?

RheingoldRiver:
I know of some wikis still using 1.35 but yeah 1.32 seems pretty safe to drop support for 👍

andrewbogott:
Please take heed that I have never worked on this codebase before and have also not tested this patch! It was written in an agitated rush once I noticed that things were breaking :)

AdamWill:
yeah, I honestly have never touched any of this code either, but it's pretty simple and matches what I would do just grepping the code for writeapi. and hey, the tests pass, so we're 70% sure it's fine! :P

AdamWill:
OK, since nobody has any objections, let's go ahead and merge this. I'm at a conference RN, I'll try and carve out more time to get back to the 0.11 release plans after that.

marcfrederick:
I've added the breaking change label to this, as it removes the public `Site.writeapi` attribute.

PR #348, opened 2024-08-12
Increase the minimum Python version to 3.6 and update syntax

marcfrederick:
This PR updates the minimum supported Python version from 3.5 to 3.6 and updates the syntax accordingly. Mainly this means using f-string for string formatting, instead of C-style formatting and the `str.format()` method.

The changes were made using `pyupgrade --py36-plus **/*.py`, followed by some manual editing and formatting.

---

~This commit updates some outdated syntax with the modern alternatives. The changes were applied automatically using pyupgrade (v2.0.1) with the following command: `pyupgrade --py3-plus **/*.py`. These updates are purely syntactical and there should be no change in functionality.~


AdamWill:
Thanks!

If we're gonna modernize string formatting I would kinda prefer to just go all the way to f-strings. They came in with Python 3.6 and are waaaay nicer than `.format()` in 99% of cases.

RHEL 7 is basically EOL, so the oldest currently supported RHEL is 8, which has Python 3.6 as its system Python. The oldest supported Debian (Buster) has Python 3.7. So I think it should be safe to go to f-strings.

wdyt?

AdamWill:
To get f-strings we'd use `pyupgrade --py36-plus` , I guess that might get us other nice 3.6+ stuff.

marcfrederick:
Yep, that would be even nicer. It would also allow us to use variable annotations ([PEP 526](https://peps.python.org/pep-0526/)) instead of using on comments in #336.

AdamWill:
Yeah, indeed. We'd have to drop 3.5 from the list in `setup.py` (and anywhere else we say we're compatible with it) but I don't really see any other consequence.

marcfrederick:
I've updated this PR to increase the minimum Python version to 3.6 and replaced `str.format()` with f-strings. @mwclient/maintainers, does anyone have any concerns about raising the minimum version?

mwclient/client.py
```diff
@@ -329,9 +327,10 @@ def api(self, action, http_method='POST', *args, **kwargs):
             >>> result = site.api('query', prop='coordinates', titles='Oslo|Copenhagen')
             >>> for page in result['query']['pages'].values():
             ...     if 'coordinates' in page:
-            ...         print('{} {} {}'.format(page['title'],
-            ...             page['coordinates'][0]['lat'],
-            ...             page['coordinates'][0]['lon']))
+            ...         title = page['title']
+            ...         lat = page['coordinates'][0]['lat']
+            ...         lon = page['coordinates'][0]['lon']
+            ...         print(f'{title} {lat} {lon}')
```
AdamWill, line 333: jesus this thing is smart

marcfrederick: Not that smart sadly, that one was me grepping manually 😅

mwclient/client.py
```diff
@@ -497,8 +494,8 @@ def raw_call(self, script, data, files=None, retry_on_error=True, http_method='P
                 stream = self.connection.request(http_method, url, **args)
                 if stream.headers.get('x-database-lag'):
                     wait_time = int(stream.headers.get('retry-after'))
-                    log.warning('Database lag exceeds max lag. '
-                                'Waiting for {} seconds'.format(wait_time))
+                    log.warning(f'Database lag exceeds max lag. '
+                                f'Waiting for {wait_time} seconds')
```
AdamWill, line 498: istr that it's technically wrong to use `.format` *or* f-strings when using `logging`; you should always do it like `log.warning('lala %s', wait_time)` (which doesn't actually 'resolve' it at that time; it's done later elsewhere). there's some background on this at https://stackoverflow.com/questions/75715376/set-the-formatting-style-for-logging-percent-format-style-bracket-or-f-stri for e.g. so perhaps for all the cases of `log.*` we should make this pr use percent instead?

AdamWill: see also https://docs.python.org/3/howto/logging-cookbook.html#using-particular-formatting-styles-throughout-your-application

RheingoldRiver:
there are no major problems with dropping support for 3.5, 6, or 7 that I know of, but if we drop support for 3.8 at any point in the near future we risk issues with sublimetext's mediawiker so I will want to make sure we stay compatible with 3.8 for a long time

I think we already have lost sublimetext 3.* with loss of python 3.3 support so that is not a concern for me

mwclient/client.py
marcfrederick: Yep good catch, I've updated the two relevant locations.

AdamWill:
That makes sense. FWIW, as a Red Hatter I'd kinda default to making the system Python version in the last supported RHEL the baseline, which means 3.6. But really, I doubt anyone actually wants to run mwclient on RHEL 8, so if there was some compelling reason to go to 3.8 it'd probably be fine.

RheingoldRiver:
nah i don't think we need to go to 3.8, I just don't want to go *past* 3.8

examples/basic_edit.py
```diff
@@ -44,10 +44,10 @@
 print('External links:', list(page.extlinks()))
 
 print('Uploading image')
-site.upload(open('test-image.png', 'rb'), prefix + '-test-image.png', 'desc', ignore=True)
+site.upload(open('test-image.png', 'rb'), f'{prefix}-test-image.png', 'desc', ignore=True)
 print('Uploading image for the second time')
-site.upload(open('test-image.png', 'rb'), prefix + '-test-image.png', 'desc', ignore=True)
-image = site.Images[prefix + '-test-image.png']
+site.upload(open('test-image.png', 'rb'), f'{prefix}-test-image.png', 'desc', ignore=True)
```
AdamWill, line 49: while we're modernizing, I think these `open` calls ought to have an explicit `encoding` arg, right?

marcfrederick: That, and `open()` should probably use a context manager, so this should be something like this instead imo:

```suggestion
with open('test-image.png', 'rb', encoding='utf-8') as f:
    site.upload(f, f'{prefix}-test-image.png', 'desc', ignore=True)
```

test/test_client.py
```diff
@@ -24,10 +24,14 @@
...
+            .format(version=kwargs.get('version', '1.24wmf17'),
+                    rights=json.dumps(kwargs.get('rights', [
+                        "createaccount", "read", "edit", "createpage", "createtalk",
+                        "editmyusercss", "editmyuserjs", "viewmywatchlist",
+                        "editmywatchlist", "viewmyprivateinfo", "editmyprivateinfo",
+                        "editmyoptions", "centralauth-merge", "abusefilter-view",
+                        "abusefilter-log", "translate", "vipsscaler-test", "upload"])))
```
AdamWill, line 35: AGH GOD MY EYES

is this...actually better? can we manually redo it better...somehow?

marcfrederick: How about something like this instead? Still not necessarily _nice_, but much more readable. I've also dropped a bunch of non-default namespaces to make it more manageable.

```python
def metaResponse(self, version='1.24wmf17', rights=None):
    if rights is None:
        rights = [
            "createaccount", "read", "edit", "createpage", "createtalk",
            "editmyusercss", "editmyuserjs", "viewmywatchlist",
            "editmywatchlist", "viewmyprivateinfo", "editmyprivateinfo",
            "editmyoptions", "centralauth-merge", "abusefilter-view",
            "abusefilter-log", "translate", "vipsscaler-test", "upload"
        ]

    # @formatter:off
    namespaces = {
        -2: {"id": -2, "*": "Media", "canonical": "Media", "case": "first-letter"},
        -1: {"id": -1, "*": "Special", "canonical": "Special", "case": "first-letter"},
        0: {"id": 0, "*": "", "case": "first-letter", "content": ""},
        1: {"id": 1, "*": "Talk", "canonical": "Talk", "case": "first-letter", "subpages": ""},
        2: {"id": 2, "*": "User", "canonical": "User", "case": "first-letter", "subpages": ""},
        3: {"id": 3, "*": "User talk", "canonical": "User talk", "case": "first-letter", "subpages": ""},
        4: {"id": 4, "*": "Wikipedia", "canonical": "Project", "case": "first-letter", "subpages": ""},
        5: {"id": 5, "*": "Wikipedia talk", "canonical": "Project talk", "case": "first-letter", "subpages": ""},
        6: {"id": 6, "*": "File", "canonical": "File", "case": "first-letter"},
        7: {"id": 7, "*": "File talk", "canonical": "File talk", "case": "first-letter", "subpages": ""},
        8: {"id": 8, "*": "MediaWiki", "canonical": "MediaWiki", "case": "first-letter", "subpages": ""},
        9: {"id": 9, "*": "MediaWiki talk", "canonical": "MediaWiki talk", "case": "first-letter", "subpages": ""},
        10: {"id": 10, "*": "Template", "canonical": "Template", "case": "first-letter", "subpages": ""},
        11: {"id": 11, "*": "Template talk", "canonical": "Template talk", "case": "first-letter", "subpages": ""},
        12: {"id": 12, "*": "Help", "canonical": "Help", "case": "first-letter", "subpages": ""},
        13: {"id": 13, "*": "Help talk", "canonical": "Help talk", "case": "first-letter", "subpages": ""},
        14: {"id": 14, "*": "Category", "canonical": "Category", "case": "first-letter"},
        15: {"id": 15, "*": "Category talk", "canonical": "Category talk", "case": "first-letter", "subpages": ""},
    }
    # @formatter:on

    return {
        "query": {
            "general": {
                "generator": f"MediaWiki {version}"
            },
            "namespaces": namespaces,
            "userinfo": {
                "anon": "",
                "groups": ["*"],
                "id": 0,
                "name": "127.0.0.1",
                "rights": rights
            }
        }
    }
```

AdamWill: yeah, I like that a lot more. thanks.

PR #242, opened 2019-11-21
Add support for clientlogin

DylannCordel:
From [the doc](https://www.mediawiki.org/wiki/API:Login)

> As of MediaWiki 1.27, using the main account for login is not supported. Obtain credentials via Special:BotPasswords or use clientlogin method.

MediaWiki 1.34 `login` method is still working with normal account but "raises Deprecated" (send content into "warnings"). This PR adds support of clientlogin to avoid those warnings and prepare to future versions when `login` method will only work with "Bot passwords"

PR #349, opened 2024-08-21
Fix/initialize size after clientlogin

marcfrederick:
This fix (first commit) ensures that the `Site` object is properly initialized after a successful `Site.clientlogin`, matching the existing behavior of `Site.login()`.

Previously, developers needed to manually invoke `Site.site_init()` after using `clientlogin`, which could lead to unexpected issues, especially during migration from `login` to `clientlogin`.

For example, the following code works with `Site.login`:

```python
site = mwclient.Site('localhost:8080', path='/', scheme='http')
site.login('username', 'super-secret')
with open('/tmp/image.png', 'rb') as f:
    site.upload(f, 'image.png', 'Some test upload', ignore=True)
```

But when using `Site.clientlogin` without this fix, the code would fail with `mwclient.errors.InsufficientPermission: image.png`:

```python
site = mwclient.Site('localhost:8080', path='/', scheme='http')
site.clientlogin(username='username', password='super-secret')
with open('/tmp/image.png', 'rb') as f:
    site.upload(f, 'image.png', 'Some test upload', ignore=True)
```

The second commit refactors the method, removing some redundant logic.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -777,6 +777,10 @@ class Site:
                      `username` and `password`
                 See https://www.mediawiki.org/wiki/API:Login#Method_2._clientlogin
 
+        Returns:
+            bool | dict: True if login was successful, or the response if it's a
+                multi-steps login process you started.
+
         Raises:
             LoginError (mwclient.errors.LoginError): Login failed, the reason can be
                 obtained from e.code and e.info (where e is the exception object) and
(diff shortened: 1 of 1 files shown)


AdamWill:
other than that, I think it LGTM. Removing the `while True` loop scares me - it makes me wonder why it's there at all - but I really can't see how it could possibly be useful, as you say. I wondered if it was vestigial (it *used* to be useful), but no, it's been this way since it was introduced in https://github.com/mwclient/mwclient/pull/242 . Still, I think this is fine.

marcfrederick:
> I'm not sure it's worth changing the string formatting here as we're planning to go to f-strings in https://github.com/mwclient/mwclient/pull/348 . maybe leave it the same here to make rebasing with that PR simpler?

Good point, I'll update the commit.

> Removing the `while True` loop scares me - it makes me wonder why it's there at all

I'm 99% sure this is an artifact from this code being copied and adapted from the `login` method. There we have a loop and sleeper, because the API can return `Throttled`, in which case we retry. For `clientlogin`, the only possible status are `PASS`, `FAIL`, `UI`, `REDIRECT`, and `RESTART`, so we don't have that retry logic here.

AdamWill:
Oh, yeah, that makes sense.

I *really* wish we had some integration testing for this thing :/ I'd like us to fill out the test suite and add an integration test using a mediawiki container before we charge too far further ahead with development. But on the whole this LGTM. I'll give it a day in case Rheingold has feedback, though.

AdamWill:
This needs a manual rebase now I merged https://github.com/mwclient/mwclient/pull/348 .

PR #352, opened 2024-08-21
Initial container-based integration testing

AdamWill:
This adds an integration test that runs a mediawiki container, configures it enough that the API works, then runs a single test against it. We also enable this in CI, if I got it right. WIP!

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,28 @@
+---
+name: Run integration tests
+
+on: ["push", "pull_request"]
+
+jobs:
+  build:
+    runs-on: "ubuntu-latest"
+    strategy:
+      max-parallel: 4
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v5.1.1
+        with:
+          python-version: '3.12'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox tox-gh-actions
+      - name: Install podman
+        run: sudo apt-get -y install podman
+      - name: Integration test with tox
+        run: tox -e integration
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -39,8 +39,8 @@ jobs:
       - name: Test with tox
         run: tox
       - name: Produce coverage.lcov for coveralls
-        if: matrix.python-version == '3.11'
+        if: matrix.python-version == '3.12'
         run: coverage lcov
       - name: Coveralls
-        if: matrix.python-version == '3.11'
+        if: matrix.python-version == '3.12'
         uses: coverallsapp/github-action@v2
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -64,6 +64,22 @@ Alternatively, you can run the tests directly with pytest:
     $ pip install -e '.[testing]'
     $ py.test
 
+There is a container-based integration test suite which is not run by default
+as it requires docker or podman, is a little slow, and needs to do ~3G of network
+transfer when first run (to download the mediawiki container images). It is
+run as part of CI. To run it locally, make sure you have docker or podman
+installed, then with tox, do:
+
+.. code:: bash
+
+    $ tox -e integration
+
+Or with pytest, do:
+
+.. code:: bash
+
+    $ py.test test/integration.py
+
 If you would like to expand the test suite by adding more tests, please go ahead!
 
 Updating/expanding the documentation
@@ -106,7 +122,8 @@ When it is ready, push your branch to your remote:
     $ git push -u fork my-branch
 
 Then you can open a pull request on GitHub. You should see a URL to do this
-when you push your branch.
+when you push your branch. Tests will be automatically run on your pull
+request via GitHub Actions.
 
 Making a release
 ----------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,9 +2,6 @@
 requires = ["setuptools>=40.6.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
-[tool.pytest.ini_options]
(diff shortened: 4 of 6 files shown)


AdamWill:
ooooo

AdamWill:
I'll flesh this out a bit more tomorrow - create some test users, test some more stuff - but it looks like it works! Yell if you see anything horribly wrong.

AdamWill:
OK, gonna call this not WIP any more, I think it's in good enough shape now.

In *theory* we should probably pin the specific mediawiki versions we test, and use pull requests to update them. Ideally with something like dependabot to notice new mediawiki versions and automatically generate PRs. But that's more fiddly to implement, and in practice mediawiki probably moves slow enough and breaks compatibility infrequently enough that it's probably okay to do it this way. If a new mediawiki version does break us, we'll find out on the first PR after it happens.

Maybe I'll look into the dependabot approach as a follow-up if I feel really keen.

marcfrederick (approved):
This looks great; thank you for the work! Since the `docker` and `podman` CLIs are more or less the same, we could even consider making it work with both by checking which of the two is available. I would assume more people already have Docker than Podman, so this would make the tests easier to run locally. I'm also fine with keeping it as is, as long as it runs in CI we should be good 🤷‍♂️.

```diff
diff --git a/test/integration.py b/test/integration.py
index dbefaa8..871f7f0 100644
--- a/test/integration.py
+++ b/test/integration.py
@@ -1,3 +1,4 @@
+import shutil
 import subprocess
 import time
 from urllib.error import URLError, HTTPError
@@ -6,6 +7,15 @@ from urllib.request import urlopen
 import mwclient
 import pytest
 
+
+_CONTAINER_RUNTIME = None
+if shutil.which("podman"):
+    _CONTAINER_RUNTIME = "podman"
+elif shutil.which("docker"):
+    _CONTAINER_RUNTIME = "docker"
+else:
+    raise RuntimeError("Neither podman nor docker is installed")
+
 @pytest.fixture(
     scope="class",
     params=[("latest", "5002"), ("legacy", "5003"), ("lts", "5004")],
@@ -23,40 +33,40 @@ def site(request):
     container = f"mwclient-{tag}"
     # create the container, using upstream's official image. see
     # https://hub.docker.com/_/mediawiki
-    args = ("podman", "run", "--name", container, "-p", f"{port}:80",
+    args = (_CONTAINER_RUNTIME, "run", "--name", container, "-p", f"{port}:80",
             "-d", f"docker.io/library/mediawiki:{tag}")
     subprocess.run(args)
     # configure the wiki far enough that we can use the API. if you
     # use this interactively the CSS doesn't work, I don't know why,
     # don't think it really matters
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "php", "/var/www/html/maintenance/install.php", "--server",
             f"http://localhost:{port}", "--dbtype", "sqlite", "--pass", "weakpassword",
             "--dbpath", "/var/www/data", "mwclient-test-wiki", "root")
     subprocess.run(args)
     # create a regular user
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "php", "/var/www/html/maintenance/createAndPromote.php", "testuser",
             "weakpassword")
     subprocess.run(args)
     # create an admin user
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "php", "/var/www/html/maintenance/createAndPromote.php", "sysop",
             "weakpassword", "--bureaucrat", "--sysop", "--interface-admin")
     subprocess.run(args)
     # create a bot user
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "php", "/var/www/html/maintenance/createAndPromote.php", "testbot",
             "weakpassword", "--bot")
     subprocess.run(args)
     # disable anonymous editing (we can't use redirection via podman
     # exec for some reason, so we use sed)
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "sed", "-i", r"$ a\$wgGroupPermissions['*']['edit'] = false;",
             "/var/www/html/LocalSettings.php")
     subprocess.run(args)
     # allow editing by users
-    args = ("podman", "exec", container, "runuser", "-u", "www-data", "--",
+    args = (_CONTAINER_RUNTIME, "exec", container, "runuser", "-u", "www-data", "--",
             "sed", "-i", r"$ a\$wgGroupPermissions['user']['edit'] = true;",
             "/var/www/html/LocalSettings.php")
     subprocess.run(args)
@@ -75,9 +85,9 @@ def site(request):
     # set up mwclient.site instance and yield it
     yield mwclient.Site(f"localhost:{port}", path="/", scheme="http", force_login=False)
     # -t=0 just hard stops it immediately, saves time
-    args = ("podman", "stop", "-t=0", container)
+    args = (_CONTAINER_RUNTIME, "stop", "-t=0", container)
     subprocess.run(args)
-    args = ("podman", "rm", container)
+    args = (_CONTAINER_RUNTIME, "rm", container)
     subprocess.run(args)
```

As a side-note, I've made good experiences with the Testcontainers library in Java/Golang for exactly this use-case, though it does not work with Podman unless you set some environment variables afaik (`DOCKER_HOST=unix:///run/user/$uid/podman/podman.sock`, `TESTCONTAINERS_RYUK_DISABLED=true`).

AdamWill:
Yeah, I guess we could do that. I'll update it.

I did look at https://github.com/avast/pytest-docker , but it didn't really seem to be giving us much compared to just doing it directly like this. If you need to run more than one container it's probably handy.

PR #356, opened 2024-08-24
Remove deprecated pytest-runner and setup.py test

marcfrederick:
The `setup.py test` command and pytest-runner have been deprecated for several years and are no longer recommended. The pytest documentation no longer recommends pytest-runner and their own documentation recommends against it.

- https://github.com/pypa/setuptools/issues/1684
- https://github.com/pytest-dev/pytest/issues/5534
- https://github.com/pytest-dev/pytest-runner/issues/50

Changes made:

- Removed pytest-runner: Eliminated the use of pytest-runner from `setup.py`.
- Moved test dependencies to the `extra-dependencies`, which can be used by tox as well as for manual installation.
- Updated the documentation to recommend using tox as the default test runner. Alternatively, users can manually run tests with the `py.test` command.

These changes will also make an eventual migration to `pyproject.toml` easier, as the `extras_require` can be copied to the `project.optional-dependencies` section.

### DIFF
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -38,33 +38,31 @@ Create a new branch for your changes:
 Test suite
 ----------
 
-mwclient ships with a test suite based on `pytest <https://pytest.org>`_.
-While it's far from complete, it can sometimes alert you if you break things.
+mwclient ships with a test suite based on `pytest <https://pytest.org>`_. While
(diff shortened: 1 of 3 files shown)


AdamWill:
LGTM, thanks. I was intending to get to this after extending the test coverage, but let's do it.

PR #357, opened 2024-08-28
100% test coverage for listing.py

AdamWill:
These commits together get listing.py test coverage up to 100%. They include a couple of functional changes where I think it makes more sense to change the code than change the tests. Particularly notably, it would bump our baseline version of mediawiki to 1.26, which was released in 2015 and has been obsolete since 2016. We could do some follow-up commits to drop conditionals for older releases, if we agree this bump is reasonable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ mwclient is a lightweight Python client library to the
 [MediaWiki API](https://mediawiki.org/wiki/API)
 which provides access to most API functionality.
 It works with Python 3.6 and above,
-and supports MediaWiki 1.16 and above.
+and supports MediaWiki 1.21 and above.
 For functions not available in the current MediaWiki,
 a `MediaWikiVersionError` is raised.
 
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -340,6 +340,8 @@ class Site:
         """
         kwargs.update(args)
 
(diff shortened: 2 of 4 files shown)


AdamWill:
I'll probably do a follow-up to extend the integration tests to cover more of the listing.py logic in the 'real world' as well, it seems like a wise precaution since we're kinda fuzzing a lot of stuff with fake responses here.

marcfrederick (approved):
For context, according to [W3Techs](https://w3techs.com/technologies/details/cm-mediawiki/1), around 86% of wikis in the real world are running MediaWiki version 1.26 or newer, while 12.6% are using versions between 1.16 and 1.25, which would be affected by this change.

That said, I'm fine with this and would even consider moving the (officially supported) minimum version to 1.27. This version is the earliest for which official Docker images are available, which would allow us to properly integration test all supported versions.

AdamWill:
OK, in the interests of conservatism, I tweaked this to keep the explicit setting of `continue=` and dropped the baseline to 1.21. We can always bump it higher later.

marcfrederick:
Seems sensible if the `&continue=` is the only thing necessary for that, thank you!

AdamWill:
otherwise does this look good?

marcfrederick:
Yeah this looks great to me, thank you!

PR #359, opened 2024-09-04
Update version_tuple_from_generator to match PHPs' rules

marcfrederick:
[In the MediaWiki documentation](https://www.mediawiki.org/wiki/Manual:$wgVersion#Example_code), versions are compared using [PHP's version_compare](https://www.php.net/manual/de/function.version-compare.php) function, which canonicalizes version strings by:

- Replacing _, -, and + with dots (.)
- Inserting dots before and after any non-numeric characters

The version string is then split on the dots, and the resulting segments are compared from left to right. This PR updates `Site.version_tuple_from_generator` to parse versions according to those rules.

Additionally, this PR:
- Adds a new check to `version_tuple_from_generator` to make sure that the first two elements in the resulting version tuple are always integers.
- Makes sure we consistently use `Site.require` instead of comparing versions manually.

---

**Note:** The new canonicalization rules and the updated check are technically breaking changes due to the function being public. However, for our internal usage nothing should change. For details, see to the test case changes in commit 3b6d83b.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -864,7 +864,7 @@ class Site:
         Raises:
             errors.APIError: A token of the given type could not be retrieved.
         """
-        if self.version is None or self.version[:2] >= (1, 24):
+        if self.version is None or self.require(1, 24, raise_error=False):
             # The 'csrf' (cross-site request forgery) token introduced in 1.24 replaces
             # the majority of older tokens, like edittoken and movetoken.
             if type not in {'watch', 'patrol', 'rollback', 'userrights', 'login'}:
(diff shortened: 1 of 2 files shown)


marcfrederick:
By the way, while working on this, I also [published a small library](https://pypi.org/project/php-version-compare/) that provides a `version_compare` function fully compatible with the one from the PHP standard library. However, I think that's overkill for our use case since we only really care about the first two segments (e.g., (1, 27)), so I decided to inline the logic here.

AdamWill:
This looks great to me. No notes. Thanks.

PR #360, opened 2024-09-05
Migrate config from setup.py to the pyproject.toml

marcfrederick:
This PR builds on #311 (thank you @NguoiDungKhongDinhDanh) and migrates all configuration from `setup.py` to `pyproject.toml`. I've also
* added extra classifiers.
* added an explicit dependency on `requests` (previously only required transitively via `requests-oauthlib`).

To keep build-compatibility with Python 3.6, I've switched the build backend from `setuptools` to `hatchling`. `setuptools` only supports `pyproject.toml` from version 61 onwards, which itself isn't compatible with Python 3.6. This does not mean we have to fully use Hatch as we can still create build using `python -m build`.

Hatchling automatically includes all non-gitignored files, so the `MANIFEST.in` file has been removed.

Once we drop Python 3.6 support, we can switch to `setuptools>=61` as the build backend if needed.


AdamWill:
This looks fine to me. Let's do it. Thanks!

PR #361, opened 2024-09-14
Various documentation updates

marcfrederick:
This PR contains various improvements to the documentation (see individual commits).

* Adds documentation to the `Page`, `Image`, and `Category` constructors.
* Adds all errors to the documentation. Previously only `mwclient.errors.InsufficientPermission` was included.
* Update the required Python/MediaWiki versions in the docs.
* Add information about the version scheme/license

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -81,7 +81,23 @@ Looking for information on contributing to mwclient development?
 MIT License
 -----------
 
-    .. include:: ../../LICENSE.md
+This project is licensed under the MIT License. The full text of the license
+is reproduced below:
+
+    .. literalinclude:: ../../LICENSE.md
+        :language: text
+
+Versioning
+----------
+
+Mwclient uses `Semantic Versioning 2.0.0 <http://semver.org>`_. Versions are
+numbered ``MAJOR.MINOR.PATCH``. The version will be incremented based on the
+following:
+
+* `MAJOR` version is incremented when backwards incompatible changes are made.
+* `MINOR` version is incremented when functionality is added in a backwards
+  compatible manner.
+* `PATCH` version is incremented when backwards compatible bug fixes are made.
 
 Indices and tables
 ------------------
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -6,9 +6,9 @@ mwclient: lightweight MediaWiki client
    :align: right
    :width: 30%
 
-Mwclient is a :ref:`MIT licensed <license>` client library to the `MediaWiki API`_
-that should work well with both Wikimedia wikis and other wikis running
-MediaWiki 1.16 or above. It works with Python 2.7 and 3.3+.
+Mwclient is an :ref:`MIT licensed <license>` client library for the `MediaWiki API`_
+that works well with both Wikimedia wikis and other wikis running
+MediaWiki 1.21 or above. It is compatible with Python 3.6 and above.
 
 .. _install:
 
--- a/mwclient/errors.py
+++ b/mwclient/errors.py
@@ -104,10 +104,6 @@ class NoSpecifiedEmail(EmailError):
     pass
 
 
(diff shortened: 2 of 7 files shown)


docs/source/index.rst
```diff
@@ -55,7 +55,7 @@ mwclient.
 Reference guide
 ---------------
 
-If you are looking for information on a specific function, class or method,
```
AdamWill, line 58: oh boy, we're having opinions about the Oxford comma now?! this could get controversial. :D

mwclient/errors.py
```diff
@@ -23,18 +34,29 @@ def __init__(self, code, info, kwargs):
...
 class EditError(MwClientError):
+    """Base class for errors related to editing pages."""
     pass
 
 
 class ProtectedPageError(EditError, InsufficientPermission):
+    """Raised when attempting to edit a protected page.
+
+    Attributes:
+        page (mwclient.page.Page): The page that was attempted to be edited.
```
AdamWill, line 56: ugh, that's clunky. "The page for which the edit attempt was made"?

mwclient/errors.py
```diff
@@ -65,6 +90,13 @@ def __str__(self):
 
 
 class LoginError(MwClientError):
+    """Base class for login errors.
+
+    Attributes:
+        site (mwclient.site.Site): The site object that the login was attempted on.
```
AdamWill, line 96: "The site object on which the login attempt was made"?

mwclient/errors.py
```diff
@@ -80,11 +112,18 @@ def __str__(self):
 
 
 class OAuthAuthorizationError(LoginError):
+    """Raised when OAuth authorization fails.
+
+    Attributes:
+        site (mwclient.site.Site): The site object that the login was attempted on.
```
AdamWill, line 118: as above

mwclient/errors.py
```diff
@@ -97,18 +136,26 @@ def __str__(self):
 
 
 class EmailError(MwClientError):
+    """Base class for email errors."""
     pass
 
 
 class NoSpecifiedEmail(EmailError):
+    """Raised when trying to email a user, that has no email specified."""
```
AdamWill, line 144: the comma seems superfluous. I might say "which" or "who" rather than "that".

mwclient/errors.py
```diff
@@ -97,18 +136,26 @@ def __str__(self):
...
     pass
 
 
 class NoWriteApi(MwClientError):
+    """The wiki does not have the write API enabled."""
```
AdamWill, line 149: this exception is actually never raised any more; all use of it was removed in cd4830a5d10ea452af73a2c098f92222b8145df8 , but we didn't remove the exception itself. maybe remove it in a preparatory commit, rather than documenting it?

AdamWill:
Aside from the minor notes this looks good! Thanks. Makes me wonder if we should have a 'build the docs' CI step, though, to check it works and maybe put a rendered version somewhere we can review? I've never done one of those before, though.

docs/source/index.rst
marcfrederick: I reverted this. Honestly, I don't have a strong opinion either way and just accepted the suggestion from my spellchecker 😅

marcfrederick:
I've updated the relevant commits with your suggestions, thanks! 

> Makes me wonder if we should have a 'build the docs' CI step, though, to check it works and maybe put a rendered version somewhere we can review? I've never done one of those before, though.

There's a [readthedocs/actions/preview@v1](https://github.com/readthedocs/actions/tree/v1/preview) action that should work for this. Do you have admin rights on the ReadTheDocs project? We would need to enable 'Preview Documentation from Pull Requests.' Afterward, it should simply be a matter of adding the action to this repository according to their docs

AdamWill:
Looks great, thanks! I don't know if any of us have admin on the readthedocs :/ I'll look into it.

PR #363, opened 2024-09-28
fix: Avoid import conflicts by renaming types.py to _types.py

marcfrederick:
Renaming `types.py` to `_types.py` to avoid conflicts with Python's built-in modules. This change addresses some `ImportErrors` such as:

```
ImportError: cannot import name 'MappingProxyType' from partially initialized module 'types' (most likely due to a circular import)
```

As far as I can tell this is not a problem when installing the library using pip or running tests in tox, only local execution when adding the source-root to the `PYTHONPATH`. This is e.g. the case in the default IntelliJ/Pycharm run configuration when testing in scratch files.


AdamWill:
Huh. That seems weird. Feels a lot like a bug in something else (python?) tbh, but maybe I'm missing something.

I do wonder if moving our mwclient under a `src` subdirectory would help with this, I've followed that pattern with my own projects for a while now.

marcfrederick:
It’s not exactly a bug; import resolution in Python is just _weird_. When module names in a local library match internal modules, and that directory is added to the `PYTHONPATH`, the local packages overwrite the internal modules. For more details see https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-name-shadowing-trap. There are tickets for this in the CPython repo, but they are all closed, e.g. python/cpython#101210.

AdamWill:
huh, interesting. never ran across that before, I must be lucky with naming things. yeah, moving it to src wouldn't help then.

PR #364, opened 2024-09-29
Allow passing `int` and `Page` objects to `PageList.get` and `__getitem__`

marcfrederick:
The previous type-annotation was too narrow here (and even required a `type: ignore[index]`). `PageList.get` also works with `int` and `Page` objects.


AdamWill:
LGTM.

PR #365, opened 2024-10-04
Update page attributes after edit

### LINKED ISSUE #63: Update Page state on save()
I find the following counter-intuitive:

```
> page = site.pages['Some new page']
> page.save('Some text')
> page.exists
False
> page.text()
''
> page.touched()
 (0, 0, 0, 0, 0, 0, 0, 0)
```

and so on.. After a save, the page does not reflect the updated state of the page. This could be solved by making extra API call to update page properties after save. For users trying to minimize the number of API calls, there should possibly be an option to disable this update..

### LINKED ISSUE #354: When you create a page, mwclient still thinks it doesn't exist
Another one I found writing integration tests. To reproduce:
```
pg = site.pages["NonExistentPage"]
pg.exists
pg.edit("Hi I'm a new page", "create new page")
pg.exists
```
`pg.exists` will be `False` both times, when obviously it should be `True` the second time (assuming the edit worked). You have to re-initialize the page after editing to get the right result:
```
pg = site.pages["NonExistentPage"]
pg.exists
pg.edit("Hi I'm a new page", "create new page")
pg = site.pages["NonExistentPage"]
pg.exists
```
The easy-looking way to fix this would be just to add a `self.exists = True` call to `Page._edit` at an appropriate point where we're really sure the edit actually *did something*, probably in the `if 'newtimestamp' in result['edit'].keys():` block. I'm not absolutely 100% sure that's safe, though. The probably-better way to fix it would be to turn `exists` into a property and have it depend on the page info...we might want to have a separate property for the page info, in fact. `Page.__init__` stashes the info it gets as `self._info` and it's never updated, which looks kind of...wrong.

Thoughts? @marcfrederick @RheingoldRiver

marcfrederick:
As discussed in #354, this change updates the page attributes without extra API calls after each (successful) call to `Page._edit` and its upstream functions (`Page.save`, `Page.edit`, `Page.append`, `Page.prepend`, and `Page.touch`). This is more or less what I suggested in [this comment](https://github.com/mwclient/mwclient/issues/354#issuecomment-2317762162).

@AdamWill [you mentioned](https://github.com/mwclient/mwclient/issues/354#issuecomment-2318077642) having an alternative patch lying around. I've checked the "Allow edits and access to secrets by maintainers" option, so you should be able to push to this branch if you have any changes/improvements over this.

resolves #354, resolves #63

### DIFF
--- a/mwclient/listing.py
+++ b/mwclient/listing.py
@@ -1,5 +1,5 @@
 from typing import (  # noqa: F401
-    Optional, Tuple, Any, Union, Iterator, Mapping, Iterable, Type
+    Optional, Tuple, Any, Union, Iterator, Mapping, Iterable, Type, Dict
 )
 
 import mwclient.image
@@ -237,7 +237,7 @@ class Category(mwclient.page.Page, GeneratorList):
         namespace: Optional[Namespace] = None
     ) -> None:
         mwclient.page.Page.__init__(self, site, name, info)
-        kwargs = {}
+        kwargs = {}  # type: Dict[str, Any]
         kwargs['gcmtitle'] = self.name
         if namespace:
             kwargs['gcmnamespace'] = namespace
--- a/mwclient/page.py
+++ b/mwclient/page.py
@@ -1,6 +1,6 @@
 import time
 from typing import (  # noqa: F401
-    Optional, Mapping, Any, cast, Dict, Union, Tuple, Iterable, List
+    Optional, Mapping, Any, cast, Dict, Union, Tuple, Iterable, List, NoReturn
 )
 
 import mwclient.errors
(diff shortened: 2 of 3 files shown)


AdamWill:
So compared to my patch, this handles the `edit` path much more extensively (nice job), but I also figured it made sense to update status in `move`:

        result = self.site.post('move', ('from', self.name), to=new_title,
                                token=self.get_token('move'), reason=reason, **data)
        if 'redirectcreated' not in result:
            # no redirect, so this page no longer exists
            self.exists = False
        return result['move']

and `delete`:

        result = self.site.post('delete', title=self.name,
                                token=self.get_token('delete'),
                                reason=reason, **data)
        # if the POST did not raise an exception, we don't exist
        self.exists = False
        return result['delete']

wdyt about those?

marcfrederick:
Yeah, makes sense. I’ve updated the PR with those changes and added some tests. I’ve also changed `Page.move` to set `Page.exists = False` when no redirect is created.

AdamWill:
@RheingoldRiver wdyt? I think you're entirely correct that the pages-as-persistent-instances design is inherently a bit wrong-headed and thus this can *always* potentially be a lie, but given we're stuck maintaining a mature library with said wrong-headed design, making it *less likely* to be a lie is overall still a win.

AdamWill (approved):
LGTM.

RheingoldRiver:
I think this should be okay, it's a pretty conservative change, and I can't imagine any sane situation in which it would break someone's code. LGTM

RheingoldRiver (approved):

PR #367, opened 2024-10-10
Replace `pkg_resources` with `importlib.metadata` for Python 3.8+

marcfrederick:
This change replaces the deprecated `pkg_resources` (part of `setuptools`) with `importlib.metadata` for Python 3.8 and above.

Switching to `importlib.metadata` allows us to drop the test-dependency on `setuptools` and gets rid of the `DeprecationWarning` raised by certain versions of `pkg_resources` when being imported:

```
DeprecationWarning: pkg_resources is deprecated as an API
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ testing = [
     "pytest-cov",
     "responses>=0.3.0",
     "responses!=0.6.0",
-    "setuptools",
+    "setuptools; python_version < '3.8'",
 ]
 
 [project.urls]
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -1,11 +1,11 @@
 import json
 import logging
+import sys
 import time
 import unittest
 import unittest.mock as mock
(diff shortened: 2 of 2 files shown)


AdamWill:
LGTM. Thanks.

AdamWill:
Needs rebasing on the pyproject.toml change.

dotlambda:
You overlooked
https://github.com/mwclient/mwclient/blob/87113730b1f41d8160057621bfd90ad88428f602/test/test_client.py#L134-L136

marcfrederick:
@dotlambda that `pkg_resources` usage is intentional for Python 3.7, which we still support. `importlib.metadata` was added to std in Python 3.8, so we only require setuptools (which contains `pkg_resources`) in older Python versions:

```
"setuptools; python_version < '3.8'"
```

It is also imported only when `sys.version_info < (3, 8)`, so it should not affect newer Python versions

PR #368, opened 2024-10-11
Hard-code ubuntu-22.04 for Python 3.7 CI jobs

marcfrederick:
ubuntu-latest is in the process of being updated from ubuntu-22.04 to ubuntu-24.04, which leads to issues with py37:

```
Version 3.7 was not found in the local cache
Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.
```

This commit hard-codes the ubuntu-version for py37 to ubuntu-22.04, similar to what we've done for py36.

See actions/runner-images#10636 for more information.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -21,6 +21,8 @@ jobs:
           - "3.13.0-alpha - 3.13"
         include:
           - os: "ubuntu-latest"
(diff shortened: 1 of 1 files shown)


AdamWill:
thanks for the fix, looks good.

Issue #340, opened 2024-07-09
Error in tests: "__main__.py: error: unrecognized arguments: --cov"


vitaly-zdanevich:
Full log https://935515.bugs.gentoo.org/attachment.cgi?id=897011

AdamWill:
well, install pytest-cov or patch pyproject.toml , I guess. Possibly we could drop the `--cov` in pyproject.toml. I copied that over from `setup.cfg` so it's not new, but generally speaking it should only really be the 'default' in CI, people wanting to run the test suite to verify functionality don't really need coverage reporting.

marcfrederick:
I've managed to reproduce this issue on both v0.10.1 and v0.11.0 using the following steps:

```
$ python -m venv .venv/
$ source .venv/bin/activate
$ git checkout v0.11.0
$ pip install pytest pytest-cov responses setuptools
$ python setup.py install
$ python -m pytest -p no:cov
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --cov
  inifile: /Users/marcfrederick/Developer/mwclient/setup.cfg
  rootdir: /Users/marcfrederick/Developer/mwclient
```

However, the current `HEAD` works fine. I've bisected the history, and starting with 96dea32, which removes the `tool.pytest.ini_options` from `pyproject.toml`, your script should work. 

The issue is your CI file explicitly prevents `pytest-cov` from being loaded (`-p no:cov`), which conflicted with our default pytest options. From the pytest help:

```
-p name               Early-load given plugin module name or entry point 
                      (multi-allowed). To avoid loading of plugins, use the 
                      `no:` prefix, e.g. `no:doctest`.
```

AdamWill:
oh, yeah, I actually had to remove it in that PR for similar reasons, I think. And it seemed kind of wrong for current standard practices anyway.

PR #369, opened 2024-10-11
test: remove pytest.ini_options from pyproject.toml

marcfrederick:
These options were removed in 96dea32 but were unintentionally reintroduced in #360 as part of 859f3ee. This can cause issues in certain CI environments (see #340). Additionally, the module name after `--cov` is incorrect, resulting in the following warning:

```
CoverageWarning: Module mwtp was never imported. (module-not-imported)
  self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")
```


AdamWill:
sigh, yeah, whoops. Sorry I missed this on review.

PR #370, opened 2024-10-16
Update `Site.expandtemplates` to use the new API added in 1.24

marcfrederick:
Refactored `Site.expandtemplates` to use the updated API, addressing deprecation warnings in recent MediaWiki versions.

The older API triggered the following warnings:

```
WARNING:mwclient.client:Subscribe to the mediawiki-api-announce mailing list at <https://lists.wikimedia.org/postorius/lists/mediawiki-api-announce.lists.wikimedia.org/> for notice of API deprecations and breaking changes. Use [[Special:ApiFeatureUsage]] to see usage of deprecated features by your application.
WARNING:mwclient.client:The parameter "generatexml" has been deprecated.
Because "prop" was not specified, a legacy format has been used for the output. This format is deprecated, and in the future the new format will always be used.
```

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -1809,21 +1809,50 @@ class Site:
             text: Wikitext to convert.
             title: Title of the page.
             generatexml: Generate the XML parse tree. Defaults to `False`.
+
+        Returns:
+            If generatexml is False, returns the expanded wikitext as a string.
+            If generatexml is True, returns a tuple of (expanded wikitext,
+            XML parse tree).
         """
-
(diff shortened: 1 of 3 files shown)


AdamWill:
oh, hey, yeah, I meant to do this after I ran into it while reviewing your types PR. Nice. Might be fun to extend the container tests to cover this also?

marcfrederick:
Good idea, I've added two tests (with/without the parsetree).

AdamWill:
Nice!

PR #372, opened 2024-10-16
docs: document why some functions accept both `*args` and `**kwargs`

### LINKED ISSUE #128: Be consistent in args vs kwargs
Glancing over the source, I often find `kwargs.update(args)`. Also, in `Listing.__init__` of `listing.py`:

``` python
        data = self.site.api(
            'query', (self.generator, self.list_name),
            *[(text_type(k), v) for k, v in six.iteritems(self.args)]
        )
```

What reason is there for this duality of 2-tuples as arguments vs keyword arguments? Why not just consistently use keyword arguments?

If there is a reason for it, documenting it might be appropriate.

marcfrederick:
Adds some documentation to explain why certain methods take both `*args` and `**kwargs` instead of just `**kwargs`. See #128 for more information.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -11,8 +11,8 @@ from requests_oauthlib import OAuth1
 
 import mwclient.errors as errors
 import mwclient.listing as listing
-from mwclient.sleep import Sleeper, Sleepers
 from mwclient._types import Cookies, Namespace, VersionTuple
+from mwclient.sleep import Sleeper, Sleepers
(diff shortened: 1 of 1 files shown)


AdamWill:
If it's not already explained somewhere, I think the information that you have to pass these as tuples would be useful (it's one of those things that obvious...*once it's been explained*).

marcfrederick:
I've updated the docstring to make this a bit more more clear. That information is also contained in the type though (`Tuple[str, Any]`), so I think it should be fairly obvious.

AdamWill:
Great.

PR #377, opened 2025-03-17
Please whitelist requesting createaccount tokens in get_token

wadoli:
Hi

According to the API documentation for MediaWiki we still seem to need the `createaccount`-tokens to add user accounts to the wiki: https://www.mediawiki.org/wiki/API:Account_creation#Python.

Could you whitelist the `createaccount` token requests, please?

Regards,
  Oliver

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -970,7 +970,14 @@ class Site:
         if self.version is None or self.require(1, 24, raise_error=False):
             # The 'csrf' (cross-site request forgery) token introduced in 1.24 replaces
             # the majority of older tokens, like edittoken and movetoken.
-            if type not in {'watch', 'patrol', 'rollback', 'userrights', 'login'}:
+            if type not in {
+                'watch',
+                'patrol',
+                'rollback',
+                'userrights',
+                'login',
+                'createaccount',
+            }:
                 type = 'csrf'
 
         if type not in self.tokens:


AdamWill:
this looks reasonable to me. @marcfrederick wdyt?

AdamWill:
oh, flake doesn't like the line length, apparently our limit is 90.

marcfrederick (changes requested):
Aside from the flake8 failure, this looks good to me. Thanks, @wadoli!

mwclient/client.py
```diff
@@ -970,8 +970,14 @@ def get_token(
...
-            if type not in {'watch', 'patrol', 'rollback', 'userrights', 'login'}:
-                type = 'csrf'
+            if type not in {
+                'watch', 
+                'patrol', 
+                'rollback', 
+                'userrights', 
+                'login', 
+                'createaccount'}:
+                    type = 'csrf'
```
marcfrederick, line 980: @wadoli, thanks again! It looks like Flake8 still has an issue with this -- something like the following should work. 

Btw, you can check it locally by running `flake8 mwclient` from the repo root (`pip install flake8` if you don't have it installed).

```suggestion
            if type not in {
                'watch',
                'patrol',
                'rollback',
                'userrights',
                'login',
                'createaccount',
            }:
                type = 'csrf'
```

AdamWill: also, once it passes, please squash to a single commit. thanks!

wadoli: Thank you for your hints and support and sorry for not reacting quicker. I was sick for the last two days. Also I have only limited git experience. I have now read and acted on https://mwclient.readthedocs.io/en/latest/development/ and everything seems to be fine.

Given the minimal change, it was quite surprising to me that my patch to lead to so much interaction. I hope the rebased commit works for you. Thank you for your good work! 👍

marcfrederick (approved):
LGTM now 🚀

PR #379, opened 2025-03-26
ci: pin third-party GitHub Actions to commit hashes

marcfrederick:
This PR pins all third-party GitHub Actions to specific commit hashes. This helps prevent supply chain attacks, like CVE-2025-30066, where a tag is updated with malicious code. Dependabot will still create PRs for new versions, but that at least gives us a chance to review changes before merging.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python
-        uses: actions/setup-python@v5.5.0
+        uses: actions/setup-python@v5
         with:
           python-version: '3.13'
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


AdamWill:
The `actions/` ones aren't really third-party, are they? They are provided directly by Github, AIUI. We already functionally trust github by...hosting on github. So I think I'd be OK with not pinning those.

Pinning the actual third-party ones (sigstore and coveralls) is probably a good idea indeed.

marcfrederick:
Hmm, I’d argue that `actions/*` workflows are still third-party code (to us), even though they are maintained by GitHub. Admittedly, the risk of compromise is significantly lower for those though, so pinning them to the major version should be fine.

AdamWill:
LGTM!

AdamWill (approved):
looks great, thanks.

PR #383, opened 2025-06-25
Drop Python 3.6 Support; Require Python >=3.7

marcfrederick:
GitHub has deprecated the `ubuntu-20.04` runner, which was the last image supporting Python 3.6: https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

Additionally, actions/setup-python will not be updated to support Python 3.6 going forward: actions/setup-python#1048

As a result, Python 3.6 is no longer easily testable and I've removed it from the matrix.

@mwclient/maintainers we should decide whether to drop support for Python 3.6 entirely (given that it's EOL for 3+ years), or at least update the documentation to mention that it's not tested in CI and only supported on a best-effort basis.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -11,7 +11,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.6"
           - "3.7"
           - "3.8"
           - "3.9"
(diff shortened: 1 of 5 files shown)


AdamWill:
I'd be OK with entirely dropping support for it, it seems easiest at this point.

PR #384, opened 2025-07-23
Update documentation for the Site constructor `pool` parameter

audiodude:
We recently had a bug where we weren't sending our custom User-Agent to the WMF servers. The source of this bug was the fact that we were using the `pool` parameter in the `Site` constructor, and therefore `clients_useragent` was being silently ignored.

In lieu of issuing warnings or throwing errors in this situation, this PR simply updates the docs so that future users will have a better chance of figuring out this situation without needing to read the code directly.


marcfrederick (approved):
LGTM, thanks for the contribution!

PR #407, opened 2025-12-23
Require custom User-Agent for wikimedia sites (#302)

AdamWill:
This requires users to set a custom User-Agent header when using Wikimedia sites, as defined by a list I populated from https://meta.wikimedia.org/wiki/Complete_list_of_Wikimedia_projects

It's a breaking change for anyone using a Wikimedia site and not currently doing this.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -20,6 +20,20 @@ __version__ = '0.11.0'
 log = logging.getLogger(__name__)
 
 USER_AGENT = f'mwclient/{__version__} (https://github.com/mwclient/mwclient)'
+WIKIMEDIA_SITES = (
+    # https://meta.wikimedia.org/wiki/Complete_list_of_Wikimedia_projects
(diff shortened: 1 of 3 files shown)


AdamWill:
Fixed, thanks.

marcfrederick:
I don't feel super strongly about this, but think client libraries are the wrong place to implement this. It requires us to keep the list of Wikimedia hosts up to date and makes mwclient behave differently depending on the site, which is pretty unexpected behavior for users (not to mention a breaking change).

I feel like the correct way to enforce this would be on the wiki side, by returning an error on API calls with no/default user agents, which we can then bubble up to the user.

RheingoldRiver:
I'm going to close this, the client library is not the correct place to implement this. But, we should make sure per https://phabricator.wikimedia.org/T417477 that we correctly forward a 403 due to invalid UA when that occurs

PR #409, opened 2026-02-13
Update OAuth docs to be user-only agnostic.

roysmith:
The current docs recommend user-only consumers and imply that those are the only kind which work.  In reality, multi-user consumers work just fine.


RheingoldRiver (approved):

RheingoldRiver:
Closes #408

Issue #55, opened 2014-07-08
Use API calls more efficiently


fhocutt:
mwclient does a lot of work through individual API calls. Consider making this more efficient by combining API calls as much as possible (e.g. using generators and combining results: <code>title=title1|title2|...</code>). One option may be a constructor method that collects Page requests and enables larger, less frequent API calls. It may be possible to take advantage of the database-like structure of the MediaWiki API and help users save bandwidth.

mitar:
One downside when you combine multiple titles together is that you can hit the limit of how large the URI in the GET request can be (because you are listing all in the query string). So if you are sending multiple titles, I suggest using POST instead.

RheingoldRiver:
There is now support in most methods for including a `with_content=True` flag to batch fetch page contents when metadata is also being fetched as part of the Listing class.

#418 is now open to cover additional cases we are missing (which I think may be just Site.categories[] but I'm not 100% sure)

I'm going to close this ticket but please add comments to #418 (or if you are reading this in the far future, make a new ticket) if there are additional optimizations you think are possible.

(Note, #416 is merged into master but has not yet been included in a release, and I don't think we are quite yet ready for a release, we need to address #418 first and possibly a few other things, and also make sure that `with_content` is documented)

AdamWill:
I'll try and pull together a release tracking ticket like last time.

PR #416, opened 2026-05-03
Support prefetching content for page lists

mszabo-wikia:
Currently operations on page lists that require accessing page content fetch the content separately for each page in the list. This is rather inefficient for common workflows such as transforming the content of a list of pages since a separate HTTP request needs to be issued to fetch the content for each page.

So, add a new optional `with_content` parameter to methods that enumerate pages and have List set appropriate parameters in this case so that the list response itself includes page contents and have `Page` look for page content in the info it was constructed with first. Add two new helpers to generate proper arguments for listing classes to avoid needing to duplicate the requisite logic across all call sites.

### DIFF
--- a/mwclient/client.py
+++ b/mwclient/client.py
@@ -1354,7 +1354,8 @@ class Site:
         generator: bool = True,
         end: Optional[str] = None,
         max_items: Optional[int] = None,
-        api_chunk_size: Optional[int] = None
+        api_chunk_size: Optional[int] = None,
+        with_content: bool = False
     ) -> listing.List:
         """
         Retrieve all pages on the wiki as a generator.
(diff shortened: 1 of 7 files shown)


RheingoldRiver:
For some context on this PR, Máté and I have been talking about this at the Wikimedia Hackathon in Milan. My hope is that the workflow of:

1. Get a list of page titles
2. Go through each one and edit/save it

should make approximately 1.02 requests per page, with the .02 including fetching protection info, metadata, and revision text.

To do this fully I think we will also need to add 2 additional things, (1) a rev_chunk_size (since page contents may exceed api result limit) and (2) an edit conflict strategy - probably represented through some enum type where the options are SKIP_WITH_LOG, FORCE_WITH_LOG, or ERROR. But this adds the core functionality and should close #55

alex4401:
LGTM, this should cut down a hefty amount of API calls for heavily ratelimited envs which has been a problem for a long time

AdamWill:
Thanks a lot for the work on this. I'll try and get to reviewing it as soon as I can. I keep not having as much time as I would like for mwclient due to $DAYJOB :(

RheingoldRiver:
I'm flying home from Florence tomorrow (was here for a couple days after the hackathon). Since Alex has also reviewed it I'd like to merge sometime next week, would you be opposed to that if you don't have a chance to look?

AdamWill:
No, I don't want to hold it up. If you're happy with it, I'm fine with merging it next week if I don't get time to look at it.

PR #419, opened 2026-05-23
Documentation copy-editing to keep examples consistant and types correct

PierreSelim:
Copy-edit Consistent naming of image variable in the files.rst doccumentation

Copy-edit typo Images -> Categories to be consistent with the example

Copy-edit proxy -> proxies to respect request signature
    proxies: dict[str, str] | None = None

Copy-edit Site.categories is not :meth: but :attr:

@AdamWill or @waldyrious if you can have a look at this, I think it improves the documentation (we can do better on the Pages, Images, Categories compatibility generator but that's a start)

### DIFF
--- a/docs/source/user/connecting.rst
+++ b/docs/source/user/connecting.rst
@@ -87,7 +87,7 @@ using the `connection_options` parameter of the :class:`~mwclient.client.Site`.
       'http': 'http://10.10.1.10:3128',
       'https': 'http://10.10.1.10:1080',
     }
-    site = mwclient.Site('en.wikipedia.org', connection_options={"proxy": proxies})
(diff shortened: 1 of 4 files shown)


AdamWill:
The changes look good to me, but the commit message needs some work, ironically. This is how you spell "consistent". This is how you spell "documentation". There's no need to keep repeating "Copy-edit". Please change it to this:
```
Documentation copy-editing to keep examples consistent and types correct

Consistent naming of image variable in the files.rst documentation
Correct typo Images -> Categories to be consistent with the example
Correct proxy -> proxies to respect request signature
    proxies: dict[str, str] | None = None
Correct Site.categories from :meth: to :attr:
```

PierreSelim:
Thanks @AdamWill, I'll update the commit message.

PR #425, opened 2026-08-11
feat: add Site.category_members

### LINKED ISSUE #418: Add support for with_content param to generators that are missing it
After #416, most methods (such as `Page.embeddedin()`) that fetch lists of pages are able to support `with_content`. However, `Site.categories[]` (and possibly some others) are not.

My preference is to create a new method `Site.category_members()` that can support arguments with `with_content`. This would standardize access methods across the library. Another option is to set `with_content` (and possibly some other flags) inside of a Site constructor. I don't think these are mutually exclusive of each other.

w3lld1:
## Summary

- I added `Site.category_members()` as a direct, consistent entry point for MediaWiki's `categorymembers` API.
- I exposed generator, pagination, filtering, sorting, and `with_content` options through the new method.
- I documented the new access path and added regression coverage for prefetched page content.

## Validation

- `.venv/bin/python -m pytest -q` — 116 passed
- `.venv/bin/flake8 mwclient`
- `.venv/bin/mypy`
- `.venv/bin/python -m sphinx -b html docs/source docs/build/html` — completed successfully (existing documentation warnings remain)
- `git diff --check`

I deferred the container-based integration suite to CI because the project documentation notes that it requires roughly 3 GB of image downloads.

Fixes #418

### DIFF
--- a/docs/source/user/implementation-notes.rst
+++ b/docs/source/user/implementation-notes.rst
@@ -8,7 +8,7 @@ Some notable exceptions:
 
 * ``Image.imageinfo`` is the imageinfo of the latest image. Earlier versions can be fetched using ``imagehistory()``.
 * ``Site.all*``: parameter ``(ap)from`` renamed to ``start``
-* ``categorymembers`` is implemented as ``Category.members``
+* ``categorymembers`` is available as ``Site.category_members`` or ``Category.members``
 * ``deletedrevs`` is ``deletedrevisions``
 * ``usercontribs`` is ``usercontributions``
 * First parameters of ``search`` and ``usercontributions`` are ``search`` and ``user``, respectively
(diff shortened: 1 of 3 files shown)


AdamWill:
Thanks for the contribution! Was this patch and/or commit message generated by or assisted by LLM technology? If so, can you provide details? Thanks.

@mwclient/maintainers we don't have an LLM contribution policy yet, AFAICS. We probably need one.

w3lld1:
Yes. I used LLM assistance in preparing both the patch and the commit message. I reviewed the final diff and ran the validation listed in the PR description: the full pytest suite, Flake8, mypy, the Sphinx documentation build, and `git diff --check`. The container-based integration suite was not run locally for the reason noted in the PR body.

